PR #16, opened 2015-11-01
Add badge for android-gems.com

liaohuqiu:
Added badge for android-gems: http://www.android-gems.com/lib/nickbutcher/plaid , it looks like this:

[Android Gems](http://www.android-gems.com/lib/nickbutcher/plaid)


nickbutcher:
Thank you very much for the contribution but i'd prefer not to add this.

Issue #45, opened 2015-11-24
upgrade to 1.5.0 of the gradle plugin


yogurtearl:
I tried upgrading to 1.5.0 of the gradle plugin but I got this error: 

Error:Execution failed for task ':bypass:transformNative_libsWithSyncJniLibsForRelease'.
> java.io.FileNotFoundException: /Users/......./GIT/plaid/third_party/bypass/build/intermediates/bundles/release/jni/lib/arm64-v8a/libbypass.so (No such file or directory)

nickbutcher:
Indeed :( 

I'm tracking this: https://code.google.com/p/android/issues/detail?id=193063

randhika:
Same issue..
how to solve this?

nickbutcher:
I don't have a resolution yet which is why i'm still using gradle plugin 1.3.1.  I'm tracking the above issue waiting for the Tools team to investigate.

yogurtearl:
@nickbutcher FYI, "Priority-Small" on the referenced ticket. maybe you can get them to bump up the priority

PR #53, opened 2015-12-09
Upgrade gradle plugin to 1.5.0

start141:
see #45

Because the early version of Gradle plugin can’t load .so file, so put it into a jar.
You just put .so to the default jin lib folder in the current version of Gradle plugin.


start141:
I signed it!

nickbutcher:
Awesome, thanks @start141!

Issue #47, opened 2015-11-30
API Compatibility 


hzsweers:
Forgive me if this is already mentioned elsewhere, I couldn't find any discussion in existing issues.

Since a lot of developers are looking to this project as a yard stick for how to implement material design and meaningful interaction, would you be open to PRs backporting some of the components to earlier API versions? Would be nice to see the support libraries do some of the heavy lifting.

Example: `InkPageIndicator` can be backported(ish) to API 15.

nickbutcher:
Hi, i've been meaning to write up my thinking on this so thanks for bringing this up.  I very intentionally created this with minSdk 21 as I wanted to see what I could create un-encumbered by backwards compatibility.  I know that this is unrealistic for most real apps but as a major goal of Plaid is to inspire developers I wanted free reign here.

Now that features have more or less stabilized, I'm very open to backporting.  It's not the highest priority on my backlog (persistence! tablets! cleaner code architecture!) but I would definitely accept PRs **if** they don't add significant code complexity.  As a sample app I want it to remain easy to learn from and any backport efforts will become decreasingly relevant as the market moves on.

I'd also like to break out reusable components into more focused libs or possibly into the Design Support Lib; these would definitely need to be backward compatible.

I've already seen a couple of backports of `InkPageIndicator` which is great!
- https://github.com/Commit451/InkPageIndicator
- https://github.com/DavidPacioianu/InkPageIndicator

hzsweers:
Great!

hzsweers:
Hey Nick, wanted to get your thoughts on this. I started going down the road of backporting what I could with minimal functional changes. These were the results/observations:

- I targeted API 15 compatability. Most of the compat issues could be avoided by targeting API 19 IMO.
- Even abiding by the minimal functional changes, the diff still looks immense just due to the shear amount of API-specific code. Most of it is still drop-in replacements fortunately, I'd just be uncomfortable opening a PR that big.
- Some things are harder than others to manage, particularly with XML. While the support libraries give us a lot, most of it is programmatic and doesn't take care of non-theme/style-XML for us. Things like transitions, animated vector drawables (though that's coming soon I hear), ripples, and all the various attributes don't have easy drop-in replacements. In light of this, I would go as far as saying that backporting isn't worth it unless AppCompat's next release has some serious XML magic planned.

Here's the branch: https://github.com/hzsweers/plaid/tree/z/compat

[...]

PR #62, opened 2016-02-06
Backport pt. 1

hzsweers:
As discussed in #47. This PR is intended to be minimal-to-no structural and functional changes, focusing just on compat drop-ins, API annotations, and beginning some work in gating off transitions that won't run on pre-lollipop.

This commit:
- Drops in compat* versions wherever possible, while including some local compat implementations as well.
- Target annotations on areas that will only be Lollipop+ (for future reference, no functional changes yet)
- Using appcompat theming where possible rather than material

Future work would focus around fully supporting conditional transitions and vector drawable interactions, taking advantage of android's resource directory versioning for API-dependent resources, and backporting some functionality with new logic (such as circular imageviews).

I know it's a lot, but after spending a lot of time this week trying to pull things apart in a modular way, it just seemed too much to keep track of. I would've started hitting rebase hell and risking unbuildable commits due to missed imports here or there. One approach might be to break it down into smaller components (per-package maybe?). Another idea would be to PR into another branch so that you could introduce the changes slowly. Let me know what you thinkm and sorry again for the large pull diff :/.

Testing-wise, I've been pretty thorough in testing it to make sure things look and behave the same way. I haven't been able to test designer news or product hunt though. DN's API website seems to be down, and PH requires my emailing them first and asking for permission.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -44,6 +44,7 @@ android {
         buildConfigField "String",
                 "PROCUCT_HUNT_DEVELOPER_TOKEN", "\"${product_hunt_developer_token}\""
 
+        generatedDensities = []
     }
     buildTypes {
         release {
--- a/app/src/main/java/io/plaidapp/data/api/dribbble/model/Comment.java
+++ b/app/src/main/java/io/plaidapp/data/api/dribbble/model/Comment.java
@@ -23,6 +23,7 @@ import android.widget.TextView;
 import java.util.Date;
 
 import io.plaidapp.util.DribbbleUtils;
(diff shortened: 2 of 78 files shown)


hzsweers:
I signed it!

nickbutcher:
Thanks for all of this work @hzsweers! As discussed elsewhere, please can you resubmit this to the `backport` branch i've just created. I think that it makes sense to work in a separate branch for this large a change until it has settled down.

Thanks!

hzsweers:
Will do! I'll close this for now and then re-open to that branch after rebasing next week some time.

PR #71, opened 2016-03-08
Reduced launcher icon size

konifar:
`ic_launcher` icons were not optimized.
So I optimized icons by using [ImageOptim](https://imageoptim.com/).

After optimized, the icons size reduced 21.3% on average.

imageoptim


konifar:
@googlebot I signed it!

nickbutcher:
Awesome, thanks! cc/ @romannurik

romannurik:
Sweet! I think the build tools do some optimization as well but good to be tidy!

konifar:
Thanks for merge!

PR #75, opened 2016-03-24
Fix edge case for bottom sheet not being dismissed

mattprecious:
If you drag the bottom sheet all the way to the dismiss offset, the
callback was not being notified.

This can be reproduced on master fairly reliably by keeping your finger down and dragging through the nav bar and onto the bezel.


nickbutcher:
Thanks Matt!

PR #82, opened 2016-04-06
Tickle using the correct translation

mattprecious:
Not sure if this was intentional or not, but I had a tickle occur during a Y translation animation and it got reset to 0 for one frame.


Jawnnypoo:
Maybe var x should also be renamed to `y`? :P

nickbutcher:
Thanks for the pull but @chrisbanes tells me that this is no longer needed as the invalidation tickle is now handled by `ViewCompat.offset*` calls.

I've just pushed 90ccc640284d291ea0a3a699e9d53c27137fdfa3 to remove this.

mattprecious:
Great! Thanks for the explanation.

PR #86, opened 2016-05-09
Fixes NPE when response body is null and adds visual indicator

amilcar-andrade:
- Fixes NPE when the dribble login response body is null
- Replaces Toast for a TextView

### DIFF
--- a/app/src/main/java/io/plaidapp/ui/DribbbleLogin.java
+++ b/app/src/main/java/io/plaidapp/ui/DribbbleLogin.java
@@ -59,13 +59,16 @@ import retrofit2.Retrofit;
 import retrofit2.converter.gson.GsonConverterFactory;
 
 public class DribbbleLogin extends Activity {
+    private static final String STATE_LOGIN_FAILED = "loginFailed";
 
     boolean isDismissing = false;
+    boolean isLoginFailed = false;
     private ViewGroup container;
     private TextView message;
     private Button login;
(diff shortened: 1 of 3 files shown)


amilcar-andrade:
@nickbutcher could you please review? While creating my dribble credentials I realized you can get a NPE when reading the response body also I changed the Toast for a SnackBar.

Thanks!

nickbutcher:
Thanks but not sure how a snackbar works with a dialog, could you upload a quick screen–record of this change? Thanks!

amilcar-andrade:
@nickbutcher I changed to use the SnackBar because of the `TODO` in line 156.  Attached are some videos showing the behavior using a SnackBa, Toast and showing the old behavior. Let me know which one you want to be using.

*TOAST*
toast

*SNACK*
snack

*CRASH*
crash_dribblbe

Jawnnypoo:
I prefer the crash one :P

nickbutcher:
Hmm, none of those are wonderful, especially the crash!! What about adding a TextView in the dialog below the login button reporting the failure (prob using @color/error). It could remain until you click the button again.

What do you think?

amilcar-andrade:
Sure, I can do that. I will post a video. Thanks!

amilcar-andrade:
@nickbutcher check it out...
textview

nickbutcher:
Looks great, thanks for the changes!

PR #97, opened 2016-07-03
Fix FabTransform circular reveal calculation

wasdennnoch:
This isn't really noticable at high transition speed, but slowed down you see that the transition from the fab to an activity cuts off after approx. 75% of the animation and just jumps to the non-clipped activity. With this fix the end radius gets calculated correctly. Same for the reversed transition.


wasdennnoch:
I signed it!

nickbutcher:
Good catch, thanks for the pull!

PR #101, opened 2016-07-11
Add Custom Tabs low-confidence prefetch to DribbleShot

andreban:
- Use URLs from image and from description on low confidence mayLaunchUrl


nickbutcher:
As discussed, i'd actually like to remove the link out to dribbble from this screen so will close this PR. Thanks for the contribution and sorry that it's not going to land :(

PR #104, opened 2016-07-18
Eneim/fix line spacing add

eneim:
When I use your CollapsingTextLayout without a pre-defined styling, the multiline title's lines are overlap each other. It turns out that when you have no ```lineHeightHint``` set, the ```lineSpacingAdd``` will be negative (to be correct: it is equal to  the **fourDip rounded** version of ```0 - fontHeight```). IMO, to be having multi line title every time, that value should be at least zero (so default line height is exactly the font height). So I add a constraint for it.

Please check out and give your idea.


nickbutcher:
Thanks!

Issue #117, opened 2016-09-23
if i cancel login to designer news, the maybeShowAccounts() method will be call again


Shellton:
that means that the AutoCompleteTextView will popup again after finishAfterTransition() was called
the suggestion following:
  add a condition: `&& isDismissing == false`  to maybeShowAccounts()'s conditions
but the must be another better way to fix this little bug. I find the getSharedElementEnterTransition was called again. i tried to call getSharedElementExitTransition().addListener , but it dosnt work. 
I have to date with my friend right now . I will try to study the new features after that. 
if you fix it with other better way ,please tell me, Thanks

nickbutcher:
Fixed by 0c5e0cb

PR #122, opened 2016-10-04
Fix of show accounts dropdown when transition is ended by backpressed issue #117

andremion:
Remove onTransitionEnd listener after it is called to avoid username.showDropDown() be called on end transition by backpressed


andremion:
I signed it!

PR #123, opened 2016-10-04
Fix of show accounts dropdown when transition is ended by backpressed issue #117

andremion:
emove onTransitionEnd listener after it is called to avoid username.showDropDown() be called on end transition by backpressed


nickbutcher:
Thanks for the fix!

PR #126, opened 2016-10-08
Move dependencies to main build.gradle

danielgomezrico:
Its easier to update/reuse/manage project dependencies


nickbutcher:
Thank you very much for the PR. Unfortunately i'd rather not merge this. Specifically I don't want to share the build config with the `Bypass` project as i'd rather keep that separate and as close to upstream as possible.

Given that, I think this indirection is unnecessary for just the main `app` module.

PR #127, opened 2016-10-08
FileProvider: Create files authority inside build.gradle

amilcar-andrade:
@nickbutcher please review and merge if you think these are good changes.


nickbutcher:
Thanks for the PR! I like the authority trick but don't see the advantage of the WekRef. You're never checking if `get()` returns null so this is liable to break. Could you resubmit with just 299dfdd?

amilcar-andrade:
Sure 👍

amilcar-andrade:
@nickbutcher done!

nickbutcher:
Thanks! cc/ @ianhanniballake

Issue #22, opened 2015-11-02
OOM Crash


Laimiux:
It happens on two different genymotion devices with 5.0 and 5.1.0, after I browse few grid items.

```
11-02 11:46:49.654 2492-2492/io.plaidapp E/art: Throwing OutOfMemoryError "Failed to allocate a 1555212 byte allocation with 828657 free bytes and 809KB until OOM"
11-02 11:46:49.654 2492-2492/io.plaidapp D/AndroidRuntime: Shutting down VM
11-02 11:46:49.654 2492-2492/io.plaidapp E/AndroidRuntime: FATAL EXCEPTION: main
11-02 11:46:49.654 2492-2492/io.plaidapp E/AndroidRuntime: Process: io.plaidapp, PID: 2492
11-02 11:46:49.654 2492-2492/io.plaidapp E/AndroidRuntime: java.lang.OutOfMemoryError: Failed to allocate a 1555212 byte allocation with 828657 free bytes and 809KB until OOM
11-02 11:46:49.654 2492-2492/io.plaidapp E/AndroidRuntime:     at dalvik.system.VMRuntime.newNonMovableArray(Native Method)
11-02 11:46:49.654 2492-2492/io.plaidapp E/AndroidRuntime:     at android.graphics.Bitmap.nativeCreate(Native Method)
11-02 11:46:49.654 2492-2492/io.plaidapp E/AndroidRuntime:     at android.graphics.Bitmap.createBitmap(Bitmap.java:812)
11-02 11:46:49.654 2492-2492/io.plaidapp E/AndroidRuntime:     at android.graphics.Bitmap.createBitmap(Bitmap.java:789)
11-02 11:46:49.654 2492-2492/io.plaidapp E/AndroidRuntime:     at android.graphics.Bitmap.createBitmap(Bitmap.java:756)
11-02 11:46:49.654 2492-2492/io.plaidapp E/AndroidRuntime:     at android.transition.TransitionUtils.createViewBitmap(TransitionUtils.java:165)
11-02 11:46:49.654 2492-2492/io.plaidapp E/AndroidRuntime:     at android.app.SharedElementCallback.onCaptureSharedElementSnapshot(SharedElementCallback.java:177)
11-02 11:46:49.654 2492-2492/io.plaidapp E/AndroidRuntime:     at android.app.ActivityTransitionCoordinator.captureSharedElementState(ActivityTransitionCoordinator.java:767)
[...]

PR #133, opened 2016-10-24
Mark app with large heap

Kisty:
Due to the fact this app is using images, using large heap gives it a bigger heap size, minimising OOMs (e.g. #22). It won't fix all OOMs but will help. 
https://developer.android.com/guide/topics/manifest/application-element.html#largeHeap


nickbutcher:
Thanks for the PR but i'm not ready to just give up and use `largeHeap` just yet! 😉

I have some ideas for reducing memory usage, particularly of GIFs.

PR #137, opened 2016-11-02
Removed synthetic accessors from AboutActivity

amilcar-andrade:
@nickbutcher please review. I watched the presentation,  enabled the InteliJ option and fixed the AboutActivity because I previously introduced `Activity host`


nickbutcher:
Thanks!

PR #143, opened 2016-11-16
Remove unwanted API response envelopes.

nickbutcher:
Use Retrofit & Gson magic to remove unwanted Response objects; letting
API calls just return the payload that we are interested in.

h/t @JakeWharton


app/src/main/java/io/plaidapp/data/api/DenvelopingConverter.java
```diff
@@ -0,0 +1,86 @@
...
+                try (JsonReader jsonReader = gson.newJsonReader(body.charStream())) {
+                    jsonReader.beginObject();
+                    while (jsonReader.hasNext()) {
+                        if (payloadName.equals(jsonReader.nextName())) {
+                            try {
+                                return adapter.read(jsonReader);
+                            } finally {
+                                body.close();
+                            }
+                        }
```
JakeWharton: Needs an `else` that calls `skipValue()` otherwise it will only work when it's the first property.

app/src/main/java/io/plaidapp/data/api/DenvelopingConverter.java
```diff
@@ -0,0 +1,86 @@
...
+                    }
+                }
+                return null;
```
JakeWharton: Missing `body.close()`. Hoist your try/finally to wrap both return statements.

amilcar-andrade:
Nice, the presentation was pretty nice -> https://skillsmatter.com/skillscasts/8416-making-retrofit-2-work-for-you?utm_content=buffer42b9d&utm_medium=social&utm_source=twitter.com&utm_campaign=buffer

nickbutcher:
Thanks for the comments @JakeWharton.

Issue #142, opened 2016-11-15
Layout issues when keyboard is shown


jahirfiquitiva:
If one opens a post, and the comment is automatically focused, making the keyboard appears, the layout gets "broken" and the position of items too.

Issue seems to persist even after keyboard is hidden...

**Screenshots:**

![cxlf7qjwiaawct0](https://cloud.githubusercontent.com/assets/10360816/20307275/b5aceb6c-ab0c-11e6-8dc3-5b100f7f2a0d.jpg)
![screenshot_20161115-081142](https://cloud.githubusercontent.com/assets/10360816/20307276/b5afc3fa-ab0c-11e6-8cc8-5d21fc61f667.png)
![screenshot_20161115-081148](https://cloud.githubusercontent.com/assets/10360816/20307277/b5b441d2-ab0c-11e6-961a-cba002c4255e.png)

PR #144, opened 2016-11-17
Fix content hover issue in Dribbble post.

JediBurrell:
When you'd open a Dribbble post where the height was smaller than ( screenHeight - minOffset ), the dribbble_shot_description would hover the image, and improperly place the FAB where you can't post your comment.

https://goo.gl/photos/GqevB37SrKm9ogzD6

These changes fix that:

https://goo.gl/photos/R9UeT8LkP3USN2eS7


nickbutcher:
Thanks for the fix! Partially resolves #142.

Issue #89, opened 2016-05-12
NullPointerException when i clicked number of followers


markic22:
Just wanted to inform you about one little bug in your application.  
`java.lang.RuntimeException: Unable to start activity ComponentInfo{io.plaidapp/io.plaidapp.ui.PlayerSheet}: java.lang.NullPointerException: Attempt to read from field 'int io.plaidapp.data.api.dribbble.model.User.followers_count' on a null object reference
                                                               at android.app.ActivityThread.performLaunchActivity(ActivityThread.java:3254)
                                                               at android.app.ActivityThread.handleLaunchActivity(ActivityThread.java:3350)
                                                               at android.app.ActivityThread.access$1100(ActivityThread.java:222)
                                                               at android.app.ActivityThread$H.handleMessage(ActivityThread.java:1795)
                                                               at android.os.Handler.dispatchMessage(Handler.java:102)
                                                               at android.os.Looper.loop(Looper.java:158)
                                                               at android.app.ActivityThread.main(ActivityThread.java:7229)
                                                               at java.lang.reflect.Method.invoke(Native Method)
                                                               at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1230)
                                                               at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:1120)
                                                            Caused by: java.lang.NullPointerException: Attempt to read from field 'int io.plaidapp.data.api.dribbble.model.User.followers_count' on a null object reference
                                                               at io.plaidapp.ui.PlayerSheet.onCreate(PlayerSheet.java:142)`

nickbutcher:
Thanks for reporting. Can you provide any more details on the steps to reach this state? I'm trying to figure out when the mode is in the intent extras but not the Player object? Also not sure what to do in this state!! Thanks.

markic22:
I just click on one random article, then I click on user icon and then on followers and that happens. Problem is that the player is already null in PlayerActivity, from which you trigger PlayerSheet. In DriblerShot you check if user has shots_cuont more than 0
`   
                    if (shot.user.shots_count > 0) { // legit user object
                        player.putExtra(PlayerActivity.EXTRA_PLAYER, shot.user);
                    } else {
                       .......
                    ActivityOptions options =
                            ActivityOptions.makeSceneTransitionAnimation(DribbbleShot.this,
                                    playerAvatar, getString(R.string.transition_player_avatar));
                    startActivity(player, options.toBundle());
                }
[...]

PR #145, opened 2016-11-18
Resolve Issue #142, #89, and IllegalStateException warning.

JediBurrell:
These changes fix Issue #142 by preventing the comment EditText from catching focus on creation.


app/src/main/res/layout/dribbble_enter_comment.xml
```diff
@@ -51,6 +51,7 @@
             android:layout_height="wrap_content"
             android:layout_weight="1"
             android:inputType="textCapSentences"
+            android:focusable="false"
```
ataulm: This would break keyboard focus (keyboard access won't trigger `onTouch`)

JediBurrell: @ataulm Correct. That's why in DribbbleShot.java (Line 713-721), it sets it back to true.

Worked fine when I tested it.

ataulm: It sets it back to true in onTouch, which won't be triggered for keyboard users

JediBurrell: @ataulm What do you mean?

To gain focus, you have to touch the `EditText` field.
The initial focus ( On screen keyboard, more realistically ) messes with the layout.

ataulm: Sorry, I wasn't clear 🙂 

You can focus both in touch or non-touch mode, where the former is possible if a View `focusableInTouchMode` property is set to true (as it is for EditText).

You could try this fix by modifying the `focusableInTouchMode` property for example and it should not affect hardware keyboard access.

If the error is that the soft keyboard is messing with the layout when you first open the activity, perhaps setting `android:windowSoftInputMode="stateHidden"` for this Activity in the manifest might be enough?

JediBurrell: @ataulm That makes sense, I've just pushed a new commit that instead utilizes `android:windowSoftInputMode="stateHidden"` and it works just as well, if not ( definitely ) better. :)

nickbutcher:
Thanks for the contribution but I went another way and resolved this in f1b8bd8925e98dca4c5a5b83303dc2ba3d83e3d2

PR #148, opened 2016-12-11
Remove unused code

takahirom:
I thought that I do not need this code.
And it causes confusion.


nickbutcher:
Thanks, looks like some old code from before I switched this to [`FabTransform`](https://github.com/nickbutcher/plaid/blob/master/app/src/main/java/io/plaidapp/ui/transitions/FabTransform.java).

PR #152, opened 2016-12-11
fixed Item can be opened twice

GaganGupta19:
I am using a boolean variable to check whether Activity has started or not and changing it's value according to it.


nickbutcher:
Thank you for the PR but I don't think i'll merge this. You can make this issue exhibit for any Android view and I don't think it's appropriate to manually track and block like this.

PR #170, opened 2017-03-25
Adds home screen search widget

amilcar-andrade:
Plaid already supports shortcuts for search
and for creating a designer story  but it is
only available on Android 7.1
Adding a Widget on the home screen is a good
enough backport for older Android versions

ezgif com-video-to-gif

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -154,6 +154,16 @@
                 android:resource="@xml/glide_disk_cache_path" />
         </provider>
 
+        <receiver android:name=".ui.widget.SearchAppWidget" android:label="@string/search_hint">
+            <intent-filter>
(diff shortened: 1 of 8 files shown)


amilcar-andrade:
@nickbutcher I added a search widget. Let me know what do you think. Thanks!

amilcar-andrade:
@nickbutcher also, Do you know how to force the widget title to show up on the home screen? Right now I have a TextView showing the text`Search` and the `drawable` but I feel that there should be an API to make the system pick the receiver `label` or the `appwidget-provider` property `label. It seems the receiver title shows up fine on the widget list but not on the home screen.

amilcar-andrade:
Maybe I am missing something but I can clearly
screen shot 2017-03-25 at 10 35 04 pm
 see that the TextView style is completely different.

PR #171, opened 2017-03-26
Adds shortcut widget for searching

amilcar-andrade:
Plaid already supports shortcuts for search
and for creating a designer story  but it is
only available on Android 7.1

Adding a shortcut widget on the home screen
is a good enough backport for older versions

ezgif com-video-to-gif

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -136,6 +136,18 @@
             </intent-filter>
         </activity>
 
+        <activity
+            android:name=".ui.SearchShortcutConfigureActivity"
+            android:theme="@style/Plaid.Translucent"
+            android:icon="@drawable/ic_shortcut_search"
+            android:label="@string/search"
+            android:excludeFromRecents="true"
+            android:taskAffinity="">
+            <intent-filter>
+                <action android:name="android.intent.action.CREATE_SHORTCUT" />
+            </intent-filter>
+        </activity>
+
         <service
             android:name=".data.api.designernews.PostStoryService"
             android:exported="false" />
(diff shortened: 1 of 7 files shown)


amilcar-andrade:
@nickbutcher please review this PR. The other one (https://github.com/nickbutcher/plaid/pull/170) was using a widget and I think it makes more sense to just use the a shortcut.

amilcar-andrade:
@nickbutcher I also created different commits in this PR. So it is easier to review. Thanks in advance!

nickbutcher:
Thanks for contributing this. I'm in two minds about merging; while I see the benefits, I'd really like to keep the app as lean as possible. As a demo app, a major goal is to demonstrate material UI and little else so that it's more understandable.

I'm considering moving the `minSdk` up to 23 or perhaps even beyond so this might not be needed soon.

What do you think?

nickbutcher:
Oh and in the future, perhaps open an issue to discuss adding a new feature before PRing – i'd hate for you to spend time on something that won't get merged.

amilcar-andrade:
@nickbutcher it is your app. So whatever you think it is the best should merge :) . 

 I just did it because it was annoying not to have the shortcuts on Android 5.0 - 7.0. I know 7.1 will give you the shortcuts but I just have 1 device running 7.1

amilcar-andrade:
@nickbutcher one last question. Do you want to keep the first and second commit? 

The first one it is just creating a dimension

The second one is to hide the save button inside the SearchActivity if the user is starting the activity from the shortcut. HomeActivity handles saving the search term inside onActivityResult.

nickbutcher:
Thanks for understanding; I think that it's best for the project not to merge this. Thanks again for your contribution and sorry that this one didn't work out.

amilcar-andrade:
No problem, good thing I first implemented the Search shortcut 😅  the post shortcut it is a little bit more complicated.

piotrek1543:
`As a demo app, a major goal is to demonstrate material UI and little else so that it's more understandable.

I'm considering moving the minSdk up to 23 or perhaps even beyond so this might not be needed soon.`

I wish there would be an app, which would show how to use Material Design with animations, elevation and more.. for devices with Android system version 4.x, at least from early Jelly Bean version :-D

@nickbutcher what about making branches for specific platform features like shortcuts in API 23?

nickbutcher:
I'm reluctant to commit to creating branches as i'm not sure I've got bandwidth to keep them up to date. I am however more than happy to link out to interesting forks like I have in the README.

amilcar-andrade:
@piotrek1543 just fork this repo and create your own branch.

PR #172, opened 2017-03-27
Fixes DesignerNewsLogin NPE

amilcar-andrade:
Check if getAuthedUser call is successful before getting the body
of the response in order to avoid a NPE.

Also, removes synthetic accessors from DesignerNewsLogin and re-
arranges methods by access modifiers

giphy

### DIFF
--- a/app/src/main/java/io/plaidapp/ui/DesignerNewsLogin.java
+++ b/app/src/main/java/io/plaidapp/ui/DesignerNewsLogin.java
@@ -210,17 +210,6 @@ public class DesignerNewsLogin extends Activity {
         maybeShowAccounts();
     }
 
-    private TextWatcher loginFieldWatcher = new TextWatcher() {
-        @Override public void beforeTextChanged(CharSequence s, int start, int count, int after) { }
-
-        @Override public void onTextChanged(CharSequence s, int start, int before, int count) { }
-
-        @Override
-        public void afterTextChanged(Editable s) {
-            login.setEnabled(isLoginValid());
-        }
-    };
-
     void maybeShowAccounts() {
         if (username.hasFocus()
                 && username.isAttachedToWindow()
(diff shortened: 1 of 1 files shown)


amilcar-andrade:
@nickbutcher pls review. Thanks!

app/src/main/java/io/plaidapp/ui/DesignerNewsLogin.java
```diff
@@ -210,6 +210,52 @@ public void onFocusChange(View v, boolean hasFocus) {
...
+        showLogin();
+        password.requestFocus();
+    }
+
+    void showLoggedInUser() {
+        final Call<User> authedUser = designerNewsPrefs.getApi().getAuthedUser();
+        authedUser.enqueue(new Callback<User>() {
+            @Override
+            public void onResponse(Call<User> call, Response<User> response) {
+                if (response.isSuccessful()) {
```
nickbutcher: Please can you just return early if it's not successful; I prefer this style as it prevents deep nesting. Thanks!

nickbutcher:
Thanks for the contribution! In the future would you mind using separate focused commits? i.e. 1 commit for the response success check, one for the synthetic accessor and one for the method reorted? It makes it easier to review. Thanks!

amilcar-andrade:
@nickbutcher I addressed both of your comments. I have two different commits now and I also return early

app/src/main/java/io/plaidapp/ui/DesignerNewsLogin.java
```diff
@@ -210,6 +210,53 @@ public void onFocusChange(View v, boolean hasFocus) {
         maybeShowAccounts();
     }
 
+    void showLoginFailed() {
+        Snackbar.make(container, "Log in failed", Snackbar.LENGTH_SHORT).show();
```
piotrek1543: `Log in failed`- did you mean `Login failed`? :-) 

I'm pretty sure that extracting this text to `strings.xml` with an String argument and getting it using method like `getString(R.string.login_failed, user.getName())` is a better approach to it than hardcoded string.

amilcar-andrade: Hi, that is how Nick coded this method --> https://github.com/nickbutcher/plaid/commit/2821dc91ac301daecd6587e48b2f18cb259984ac

Also, I understand you can get the string from `strings.xml` using `getString()`. Maybe it was not a big deal to hardcoded the string in this case.

@nickbutcher could you confirm.

piotrek1543: @amilcar-andrade so hardcoded string was from Nick? How could i missed it? :-D

amilcar-andrade: I just tagged you in the PR that introduced it.

nickbutcher: Happy for you to fix my sloppiness in this PR if you want!!

app/src/main/java/io/plaidapp/ui/DesignerNewsLogin.java
```diff
@@ -221,6 +221,9 @@ void showLoggedInUser() {
         authedUser.enqueue(new Callback<User>() {
             @Override
             public void onResponse(Call<User> call, Response<User> response) {
+                if (!response.isSuccessful()) {
```
nickbutcher: Can you just do this as a one line please, matches style elsewhere in the project. Thanks!

amilcar-andrade: So you want `if (!response.isSuccessful) return;` ?

piotrek1543: @nickbutcher you mean removing branches like `if (!response.isSuccessful()) return`, right?

nickbutcher: Exactly, thanks!

amilcar-andrade:
@nickbutcher done with your changes!

nickbutcher:
Thanks for the changes!

PR #173, opened 2017-03-28
Extracting hardcoded strings. Reusing them if it is possible.

piotrek1543:
Hi,

I extracted all hardcoded strings to reuse them if it would be possible. Hope you would find it useful.

Cheers!


piotrek1543:
I've already checked also Java classes and there are many hardcoded values, but as most of them is used only in that specific class, so I won't extract them as this action won't add any value to project.

piotrek1543:
@nickbutcher done. I refreshed my commit according to your suggestion and deleted all disused (after change) items from `strings.xml`.

nickbutcher:
Thanks!

PR #192, opened 2017-06-12
Fix typo in dribble_fav strings

alexjlockwood:
I didn't test this, but I am assuming it is a typo (LMK if I am mistaken)


alexjlockwood:
I signed it!

nickbutcher:
Thanks @alexjlockwood! Definitely looks like a type – the danger of coding these things by hand. We need smarter tools ;)

PR #198, opened 2017-07-28
Merge Android-O feature branch

nickbutcher:


TheCraftKid:
Hey, it doesn't appear that these changes have made their way onto the Play Store yet. What's the issue with that? Is an update imminent for release?

PR #216, opened 2017-09-26
Fix ElasticDragDismissFrameLayout bump on fast scroll

### LINKED ISSUE #185: ElasticDragDismissFrameLayout scrolling too far on fast scroll gesture
I encountered a strange Issue while working with the ElasticDragDismissFrameLayout

Everything works fine while scrolling in the content and then dragging it over its bounds (so that events having dyUnconsumed !=0 are generated).

But whenever the containing ScrollView is already scrolled all the way up or down and when I'm doing a fast scroll motion results in the ElastigDragDismissFrameLayout beein scrolled way further than it should be. The issue here is, that if one returns to the starting point of the scroll motion, the view is not at its starting position, and is dismissed when stopping the scroll gesture.

Here are three examples, rawOffsetPixels of onDrag(..) is displayed in the corner:

[https://drive.google.com/open?id=0BxJvibam28UTcER0dXZ6c3I1a0k](url)
Notice how rawOffsetPixels is reported as 3207 pixels, whereas the touch tracker from the android developer settings reports a total dY of 1761 pixels.

[https://drive.google.com/file/d/0BxJvibam28UTaTBrZW9wVURJQXc/view?usp=sharing](url)
Starting the scroll fast and returning to the starting point of the scroll, rawOffsetPixels is still about 750 pixels.

[https://drive.google.com/file/d/0BxJvibam28UTQmdGamI2SDlPRWc/view?usp=sharing](url)
When starting the scroll slower, everything works as intended

My guess what is happening is that there is a conflict between setting translationY and the reported onNestedScroll(..., dyUnconsumed). There might be a race condition, where setting the translation will result in another onNestedScroll(..) being fired, with a dyUnconsumed value higher as it should be.

While less severe, i also could find this behaviour in the Plaid app, especially when scrolling over the end at bottom. I guess this somehow depends on layout complexity or main thread activity.

I will investigate further, maybe a separate transparent ScrollView can be used that is not translated/scaled but sends the scroll events to the inner ScrollView.

castorflex:
As described in issue #185 and in [this][YT] video, this
widget behave really weirdly when the scroll is performed
quickly, like a beginning of a fling gesture.

This is due to the 'returning' animation made in onStopNestedScroll.
Indeed, this method is called by the framework when ACTION_DOWN is
performed as a 'defensive cleanup for new gestures'.

This commit fixes it by retrieving the last actionEvent from
onInterceptTouchEvent, and perform the animation only if the last
actionEvent is not a ACTION_DOWN event.

Closes #185

[YT]: https://youtu.be/t-kYvsXIDoM

### DIFF
--- a/app/src/main/java/io/plaidapp/ui/widget/ElasticDragDismissFrameLayout.java
+++ b/app/src/main/java/io/plaidapp/ui/widget/ElasticDragDismissFrameLayout.java
@@ -21,6 +21,7 @@ import android.content.Context;
 import android.content.res.TypedArray;
 import android.graphics.Color;
 import android.util.AttributeSet;
+import android.view.MotionEvent;
 import android.view.View;
 import android.widget.FrameLayout;
 
@@ -50,6 +51,7 @@ public class ElasticDragDismissFrameLayout extends FrameLayout {
     private float totalDrag;
     private boolean draggingDown = false;
     private boolean draggingUp = false;
+    private int mLastActionEvent;
 
     private List<ElasticDragDismissCallback> callbacks;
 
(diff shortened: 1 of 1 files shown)


nickbutcher:
Sorry for the slow response here but thanks for this fix!

PR #219, opened 2017-10-17
Fix bug where tall cells break subsequent rows

### LINKED ISSUE #218: SpannedGridLayoutManager doesn't lay out rows below a contiguous blocks of "tall" cells
The simplest example of this problem is given by a `SpannedGridLayoutManager` with 1 column, 1.0 aspect ratio and all cells having dimensions 1 by 2. The recycler view will be initially be filled properly, but when scrolled down, no new cells/view will be laid out. Once all drawn views have scrolled off the top of the screen a `NullPointerException` is thrown.

The same problem can occur with multiple columns as long as there's a contiguous block of "tall" cells spanning all columns, see the below example with three columns:

GIF showing problem

I have found a potential solution for this problem and will raise a PR soon.

maxspencer:
Fixes #218

### DIFF
--- a/app/src/main/java/io/plaidapp/ui/recyclerview/SpannedGridLayoutManager.java
+++ b/app/src/main/java/io/plaidapp/ui/recyclerview/SpannedGridLayoutManager.java
@@ -220,7 +220,7 @@ public class SpannedGridLayoutManager extends RecyclerView.LayoutManager {
                 scrolled = dy;
             }
             if (top - scrolled >= 0) { // new top row came on screen
-                int newRow = firstVisibleRow - 1;
+                int newRow = getPreviousSpannedRow(firstVisibleRow);
                 if (newRow >= 0) {
                     int startOffset = top - (firstVisibleRow * cellHeight);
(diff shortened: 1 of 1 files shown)


piotrek1543:
@maxspencer please accept CLA and after that add another comment with  text `I signed it!` to continue.
Otherwise, this PR would be closed without merging changes.

nickbutcher:
Closing due to inactivity. Please feel free to re-open if you decide to sign the CLA.

maxspencer:
I signed it!

maxspencer:
Hi @nickbutcher,

I'd like to publish this layout manager as a separate artifact. We use it in a couple of places in The Guardian's news app now and it was quite important in allowing us to switch to using `RecyclerView`, especially on tablet where we have some quite convoluted layouts.

As well as the fix contained in this PR we've made a couple of other changes since I made a copy of this in our codebase. It would be great to re-write in Kotlin too, which I see is something you're working on for Plaid 2.0 anyway. Thanks, Max

Issue #224, opened 2017-10-23
"Cannot call this method in a scroll callback" warning in the home screen


alexjlockwood:
Noticed this warning while messing around with the dev build:

```
10-23 06:43:00.243 21023-21023/io.plaidapp W/RecyclerView: Cannot call this method in a scroll callback. Scroll callbacks mightbe run during a measure & layout pass where you cannot change theRecyclerView data. Any method call that might change the structureof the RecyclerView or the adapter contents should be postponed tothe next frame.
                                                           java.lang.IllegalStateException:  android.support.v7.widget.RecyclerView{d993774 VFED.V... .F....ID 0,0-1440,2560 #7f0b0070 app:id/grid}, adapter:io.plaidapp.ui.FeedAdapter@da249d, layout:android.support.v7.widget.GridLayoutManager@2779d12, context:io.plaidapp.ui.HomeActivity@cb58ce3
                                                               at android.support.v7.widget.RecyclerView.assertNotInLayoutOrScroll(RecyclerView.java:2673)
                                                               at android.support.v7.widget.RecyclerView$RecyclerViewDataObserver.onItemRangeInserted(RecyclerView.java:5074)
                                                               at android.support.v7.widget.RecyclerView$AdapterDataObservable.notifyItemRangeInserted(RecyclerView.java:11564)
                                                               at android.support.v7.widget.RecyclerView$Adapter.notifyItemInserted(RecyclerView.java:6871)
                                                               at io.plaidapp.ui.FeedAdapter.dataStartedLoading(FeedAdapter.java:669)
                                                               at io.plaidapp.data.BaseDataManager.dispatchLoadingStartedCallbacks(BaseDataManager.java:143)
                                                               at io.plaidapp.data.BaseDataManager.loadStarted(BaseDataManager.java:114)
                                                               at io.plaidapp.data.DataManager.loadSource(DataManager.java:95)
                                                               at io.plaidapp.data.DataManager.loadAllDataSources(DataManager.java:60)
                                                               at io.plaidapp.ui.HomeActivity$4.onLoadMore(HomeActivity.java:189)
                                                               at io.plaidapp.ui.recyclerview.InfiniteScrollListener.onScrolled(InfiniteScrollListener.java:54)
                                                               at android.support.v7.widget.RecyclerView.dispatchOnScrolled(RecyclerView.java:4722)
                                                               at android.support.v7.widget.RecyclerView.dispatchLayoutStep3(RecyclerView.java:3789)
[...]

PR #226, opened 2017-10-25
Fixed warning in RecyclerView during infinite scrolling

fobo66:
It's a fix for issue #224 as proposed [here](https://guides.codepath.com/android/Endless-Scrolling-with-AdapterViews-and-RecyclerView#troubleshooting).

### DIFF
--- a/app/src/main/java/io/plaidapp/ui/recyclerview/InfiniteScrollListener.java
+++ b/app/src/main/java/io/plaidapp/ui/recyclerview/InfiniteScrollListener.java
@@ -51,7 +51,11 @@ public abstract class InfiniteScrollListener extends RecyclerView.OnScrollListen
         final int firstVisibleItem = layoutManager.findFirstVisibleItemPosition();
 
         if ((totalItemCount - visibleItemCount) <= (firstVisibleItem + VISIBLE_THRESHOLD)) {
-            onLoadMore();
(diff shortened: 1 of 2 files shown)


nickbutcher:
Thanks for the contribution!

I think we should fix this further down the stack as it's likely an issue elsewhere. What about moving the `post` call into `InfiniteScrollListener` i.e. to `post` the call to `onLoadMore`.

app/src/main/java/io/plaidapp/ui/recyclerview/InfiniteScrollListener.java
```diff
@@ -51,7 +51,11 @@ public void onScrolled(RecyclerView recyclerView, int dx, int dy) {
         final int firstVisibleItem = layoutManager.findFirstVisibleItemPosition();
 
         if ((totalItemCount - visibleItemCount) <= (firstVisibleItem + VISIBLE_THRESHOLD)) {
-            onLoadMore();
+            recyclerView.post(new Runnable() {
```
nickbutcher: Can you create the runnable once in the constructor rather than on every load?

fobo66: It makes sense, I'll do it.

nickbutcher:
Thanks for the contribution 🙌

PR #230, opened 2017-11-04
Fix issue with duplicate gifs playing

charlesdurham:
See https://github.com/hzsweers/CatchUp/issues/67 and https://github.com/hzsweers/CatchUp/pull/78.


nickbutcher:
Thanks for the pull but this seems to be resolved in 08a4e4d.

PR #239, opened 2017-12-21
use asl for close to down avd

alexjlockwood:
Use an AnimatedStateListDrawable for the close-to-down avd in the player sheet screen.


nickbutcher:
Much cleaner—thanks!

PR #240, opened 2017-12-21
Use bottom sheet behavior from the design library

alexjlockwood:
I couldn't test the designer news activity because I don't have a client ID/secret and access token yet... but the player sheet screen works well.


nickbutcher:
What's the motivation behind this change? After checking this change out and playing with it i'm disinclined to merge as:
- There's a jump when content loads and the sheet expands
- This loses the spring physics which slows down then reverses downward flings without sufficient velocity to dismiss
What do you think the benefits of this are?

alexjlockwood:
Feel free to close this pull request then. :)

IIRC you wrote this bottom sheet behavior before the design lib made BottomSheetBehavior public. So I figured you just hadn't gotten around to migrating it yet.

Fair point about the content loading after the sheet expands (about to take a look, but guessing it is an oversight on my part). I removed the spring physics stuff because I figured you would want plaid to be consistent with the motion physics of bottom sheet behavior, but if not then that'd be another reason to close this I suppose. :)

nickbutcher:
Yes I wrote this before the Design Lib's `BottomSheet` was a thing… but I actually prefer some of the choices (like some of the motion curves) so will close. Thanks anyway for the PR—feel free to open issues to discuss changes before spending time working on them.

alexjlockwood:
*shrugs* It was the first time I ever used BottomSheetBehavior so it was worth learning how to use it anyway. :)

PR #243, opened 2017-12-22
More improvements for player landscape

amilcar-andrade:


amilcar-andrade:
@nickbutcher I added some of your suggestions plus some small fixes

app/src/main/java/io/plaidapp/ui/PlayerActivity.java
```diff
@@ -256,7 +256,8 @@ public void onDataLoaded(List<Shot> data) {
             @Override
             public void onLayoutChange(View v, int left, int top, int right, int bottom,
                                        int oldLeft, int oldTop, int oldRight, int oldBottom) {
-                getColumnCount();
+                shots.removeOnLayoutChangeListener(this);
+                getColumnCount(right, left);
```
nickbutcher: To pass width and height shouldn't the params be `right - left` and `bottom - top`?

app/src/main/res/layout-w480dp-land/activity_dribbble_player.xml
```diff
@@ -57,110 +57,127 @@
...
-            android:paddingEnd="@dimen/padding_normal"
-            android:paddingBottom="@dimen/spacing_normal"
-            android:textAppearance="@style/TextAppearance.DarkButton"
-            android:text="@string/follow"
-            android:theme="@style/LightRippleOverlay"
-            style="@style/Widget.AppCompat.Button.Colored" />
-
-        <io.plaidapp.ui.widget.BaselineGridTextView
-            android:id="@+id/player_bio"
+        <ScrollView
```
nickbutcher: Does this introduce any clipping issues in the enter transition?

nickbutcher: Also can you drag close the screen from the bio area? Perhaps should be a `NestedScrollView`?

amilcar-andrade: Oh, good call. I did not see any clipping I will post a video shortly

app/src/main/java/io/plaidapp/ui/PlayerActivity.java
amilcar-andrade: hehe 😅 , yeah your are right

app/src/main/res/layout-w480dp-land/activity_dribbble_player.xml
amilcar-andrade: @nickbutcher here is a video --> https://youtu.be/SbY1BnMbIvU

amilcar-andrade:
@nickbutcher addressed your comments and fixup all of the commits

amilcar-andrade:
@nickbutcher any updates?

nickbutcher:
Sorry for slow reply over the holidays.

Thanks for the changes. Mostly looks good to me. I have just spotted an issue; if you leave your device in landscape and repeatedly open and close `PlayerActivity` I'm seeing it alternate between the two pane and a single pane layout?!

player-landscape-bug

Can you take a look?

app/src/main/java/io/plaidapp/ui/PlayerActivity.java
```diff
@@ -423,10 +424,7 @@ private void setFollowerCount(int count) {
         }
     }
 
-    void getColumnCount() {
-        final int width = shots.getWidth();
-        final int height = shots.getHeight();
-
+    void getColumnCount(int width, int height) {
```
nickbutcher: Perhaps `updateColumnCount` would be better as this isn't just a getter?

app/src/main/java/io/plaidapp/ui/PlayerActivity.java
```diff
@@ -445,7 +443,7 @@ void getColumnCount() {
         // here's the magic incantation that does what we want
         res.updateConfiguration(viewScopedConfig, viewScopedDM);
 
-        // resources now act as if the device is the size of this view!?! Mwahahaha.
+        // resources now act as if the device is the size of this view!!! Mwahahaha.
```
nickbutcher: perhaps replace "this view" with "the grid"?

amilcar-andrade:
@nickbutcher I can take a look, the value of the value of `isTwoColumn` is returning `false`. Which is super weird. Anyway, will take a look.

amilcar-andrade:
@nickbutcher today I finally had some time to look into the issue. Please look at the second comment. Prolly you will be able to tell me why `updateConfiguration` seems to not update immediately the values or it seems it caches the last value. I was able to fix the problem by using `createConfigurationContext`. But the problem is def with `updateConfiguration` even some layouts and values are still returning the old resources. If you remove the changes from `https://github.com/nickbutcher/plaid/commit/edbeaf5a0ca522160c02093f79f4a3bae7667520#diff-301c34f8db0413ab9ce9e78438ea9e13R446` and keep using the 2 constants then it will work as expected.

amilcar-andrade:
@nickbutcher also, if this approach is not the correct one I am happy to debug more. Have a nice day!

nickbutcher:
Thank you for the PR but we're currently working on a large change to Plaid (see README) including updating the Dribbble support to what is available in the much reduced v2 API. Closing this PR whilst this work is going on. Once things settle down we could discuss re-opening. Thanks again for the contribution!

PR #244, opened 2017-12-23
Use lambdas and method references when possible

alexjlockwood:
This CL was mostly automated using Android Studio (i.e. "Analyze" --> "Inspect code" and then I used quick fixes to replace anonymous classes with lambdas and method references).


nickbutcher:
Nice – thanks!

PR #245, opened 2018-01-18
Fix ink page indicator null pointer exception

pranavpandey:
In setSelectedPage(int) method, dotCenterX might be null on various
configuration changes like entering or exiting the multi-window mode.
This issue can be replicated by following the steps below:
1. Open about section within the Plaid app.
2. Navigate to the second page, it will not appear on the first page.
3. Try to enter in the multi-window mode and then resize the window
by draging the handle. The app will crash with null pointer exception.

This issue has been fixed and can be tested in the app below:
[Rotation - Orientation Manager](https://play.google.com/store/apps/details?id=com.pranavpandey.rotation)


pranavpandey:
I signed it!

nickbutcher:
Thanks!

PR #261, opened 2018-05-30
Fix build config typo

zmdominguez:
screen shot 2018-05-29 at 00 09 47

Caught this during Tor's talk at I/O and I can't unsee it.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -45,7 +45,7 @@ android {
         buildConfigField "String",
                 "DESIGNER_NEWS_CLIENT_SECRET", "\"${designer_news_client_secret}\""
         buildConfigField "String",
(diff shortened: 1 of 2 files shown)


Mauin:
I guess this would also require renaming the usage of `PROCUCT_HUNT_DEVELOPER_TOKEN` in `BaseDataManager` after re-generating the `BuildConfig` class. https://github.com/nickbutcher/plaid/blob/master/app/src/main/java/io/plaidapp/data/BaseDataManager.java#L164

zmdominguez:
@Mauin Yes it would.

nickbutcher:
Good spot 🕵️‍♀️! Thanks for the pull.

PR #265, opened 2018-06-05
add transition when layout changed

woyunowuyuda:
Add transition when view layout changed.It is to deal with the problem of screen flickering due to input method changes(HuaWei M10 who has security keyboard).


nickbutcher:
Thank you for the PR but unfortunately this is not something we'd like to merge into the project.

PR #292, opened 2018-06-12
Add initial config for CircleCI

tiembo:
Temporarily muting lint errors during initial CI setup.

Addresses #280

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -53,6 +53,10 @@ android {
     }
 
     dynamicFeatures = [':features:activities']
+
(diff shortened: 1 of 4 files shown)


app/build.gradle
```diff
@@ -53,6 +53,10 @@ android {
     }
 
     dynamicFeatures = [':features:activities']
+
+    lintOptions {
+        abortOnError false
```
florina-muntenescu: I would enforce this to true. What do you think?
Same for the other ones

PR #297, opened 2018-06-14
Turn on lint and fix issues in feature and third_party

tiembo:


features/activities/src/main/res/values/styles.xml
```diff
@@ -37,8 +37,6 @@
     </style>
 
     <style name="Plaid.Translucent.DesignerNewsLogin">
-        <item name="android:statusBarColor">@android:color/transparent</item>
```
florina-muntenescu: the other 2 should be removed

tiembo: Done

PR #311, opened 2018-06-18
Calling the Designer News Comments API

florina-muntenescu:


base/src/test/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepositoryTest.kt
```diff
@@ -0,0 +1,130 @@
...
+        Mockito.`when`(service.getComments("11,12"))
+                .thenReturn(CompletableDeferred(resultChildrenError))
+        var successCalled = false
+        var result = emptyList<Comment>()
+
+        // When getting the comments from the repository
+        repository.getComments(listOf(1L), { list ->
+            result = list
+            // Then the correct list is received
+            Assert.assertEquals(list.size, 1)
```
nickbutcher: import method.

nickbutcher: Remove - checked later.

base/src/test/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepositoryTest.kt
```diff
@@ -0,0 +1,130 @@
...
+            successCalled = true
```
nickbutcher: Remove.

base/src/main/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,103 @@
...
+        children: List<Comment>
+    ): List<Comment> {
+        children.map { child ->
+            parents.filter { parent -> parent.id == child.links.parentComment }
+                    .map { parent -> parent.addComment(child) }
+        }
+        return parents
+    }
+
+    companion object {
```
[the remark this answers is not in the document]
florina-muntenescu: to be done in a separate PR.

base/src/main/java/io/plaidapp/base/designernews/data/api/model/Comment.java
```diff
@@ -130,12 +154,50 @@ public Builder setUserJob(String user_job) {
...
+
         public Comment build() {
             return new Comment(id, body, body_html, created_at, depth, vote_count, user_id,
-                    user_display_name, user_portrait_url, user_job, null);
+                    user_display_name, user_portrait_url, user_job, commentLinks,
+                    comments);
         }
     }
 
+    @Override
```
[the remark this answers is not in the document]
florina-muntenescu: not in the PR, as this class is used in A LOT of places

base/src/test/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRemoteDataSourceTest.kt
```diff
@@ -0,0 +1,105 @@
...
+
+    @Test
+    fun getComments_whenRequestSuccessful() {
+        // Given that the service responds with success
+        val result = Response.success(childrenComments)
+        Mockito.`when`(service.getComments("1")).thenReturn(CompletableDeferred(result))
+
+        runBlocking {
+            // When getting the list of comments
+            val response = dataSource.getComments(listOf(1L)).await()
```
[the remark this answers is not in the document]
florina-muntenescu: here it doesn't matter. but will update in the places where it does

base/src/test/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRemoteDataSourceTest.kt
```diff
@@ -0,0 +1,105 @@
...
+                assertEquals(childrenComments[i], response?.get(i)!!)
+            }
+        }
+    }
+
+    @Test
+    fun getComments_forMultipleComments_whenRequestSuccessful() {
+        // Given that the service responds with success for specific ids
+        val result = Response.success(childrenComments)
+        Mockito.`when`(service.getComments("1,2")).thenReturn(CompletableDeferred(result))
```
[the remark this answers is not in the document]
florina-muntenescu: these ids don't matter for the test data.

base/src/test/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRemoteDataSourceTest.kt
```diff
@@ -0,0 +1,105 @@
...
+import org.junit.Test
+import org.mockito.Mockito
+import retrofit2.Response
+
+/**
+ * Tests for [DesignerNewsCommentsRemoteDataSource] that mock the Designer News API
+ */
+class DesignerNewsCommentsRemoteDataSourceTest {
+
+    private val service = Mockito.mock(DesignerNewsService::class.java)
```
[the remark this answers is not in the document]
florina-muntenescu: there's only one method being called here in all tests: `service.getComments`. If the method is called with other params, then the test will fail.

base/src/test/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRemoteDataSourceTest.kt
```diff
@@ -0,0 +1,105 @@
...
+    private val dataSource = DesignerNewsCommentsRemoteDataSource(service)
+
```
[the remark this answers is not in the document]
florina-muntenescu: that's what `getComments_forMultipleComments_whenRequestSuccessful` is testing. I'll rename that one.

base/src/test/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepositoryTest.kt
```diff
@@ -0,0 +1,130 @@
...
+        repository.getComments(listOf(11L), { Assert.fail() }, { errorCalled = true })
+
+        // Then the error lambda is called
+        assertTrue(errorCalled)
+    }
+
+    @Test
+    fun getComments_multipleReplies_whenRequestSuccessful() {
+        // Given that:
+        // When requesting childrenComments for ids 1 from service we get the parent comment but
```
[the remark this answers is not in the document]
florina-muntenescu: we get the reply ids, but we don't get the replies embedded. comment updated

base/src/main/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,107 @@
...
+
+import io.plaidapp.base.designernews.data.api.model.Comment
+import kotlinx.coroutines.experimental.CommonPool
+import kotlinx.coroutines.experimental.android.UI
+import kotlinx.coroutines.experimental.launch
+import kotlinx.coroutines.experimental.withContext
+import kotlin.coroutines.experimental.CoroutineContext
+
+/**
+ * Repository for Designer News comments. Works with the service to get the data.
```
nickbutcher: Works with the remote data source?

base/src/main/java/io/plaidapp/base/designernews/data/api/model/Comment.java
```diff
@@ -65,9 +71,25 @@ public Comment(long id,
...
+    public CommentLinks getLinks() {
+        return links;
+    }
+
+    public List<Comment> getComments() {
+        return comments;
+    }
+
+    public void addComment(Comment comment) {
+        if(comments == null){
```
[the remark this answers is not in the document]
nickbutcher: Not addressed

base/src/main/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,107 @@
...
+ */
+class DesignerNewsCommentsRepository(
+    private val remoteDataSource: DesignerNewsCommentsRemoteDataSource,
+    private val uiContext: CoroutineContext = UI,
+    private val ioContext: CoroutineContext = CommonPool
+) {
+
+    /**
+     * Gets comments, together will all the replies from the API. The result is delivered to
+     * [onSuccess] and the error to [onError], on the [uiContext].
```
nickbutcher: nit: *or* the error to…

base/src/test/java/io/plaidapp/base/designernews/data/api/comments/CommentsTestData.kt
```diff
@@ -0,0 +1,61 @@
...
+
+val reply2 = Comment.Builder()
+        .setId(12L)
+        .setCommentLinks(links)
+        .setBody("commenty comment")
+        .build()
+
+val replies = listOf(reply1, reply2)
+
+val parentLinks = CommentLinks("userid", "storyid", 1L, arrayListOf(11L, 12L))
```
nickbutcher: named args as well?

base/src/test/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRemoteDataSourceTest.kt
```diff
@@ -0,0 +1,107 @@
...
+        Mockito.`when`(service.getComments("1")).thenReturn(CompletableDeferred(result))
+
+        runBlocking {
+            // When getting the list of comments
+            val response = dataSource.getComments(listOf(1L)).await()
+
+            // Then the response is the expected one
+            assertNotNull(response)
+            assertEquals(replies.size, response?.size)
+            replies.forEachIndexed { index,
```
nickbutcher: Should we just compare the lists of comments rather than looping over them?

florina-muntenescu: so much java habits facepalm

base/src/main/java/io/plaidapp/base/designernews/data/api/model/Comment.java
```diff
@@ -65,7 +73,23 @@ public Comment(long id,
...
+    public CommentLinks getLinks() {
+        return links;
+    }
+
+    public List<Comment> getReplies() {
+        return replies;
+    }
+
+    public void addReply(Comment comment) {
+        if(replies == null){
```
florina-muntenescu: @Nick - Retrofit seems to always return `replies` as null, since it's not a field, even if I mark it as `transient`. 
No idea what to do here. 
Also, in a next iteration on this code, the API response object will be separated from the business object that is used in the app.

base/src/main/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRemoteDataSource.kt
```diff
@@ -0,0 +1,63 @@
...
+    private val service: DesignerNewsService,
+    private val ioContext: CoroutineContext = CommonPool
+) {
+
+    /**
+     * Get a list of comments based on ids from Designer News API.
+     * If the response is not successful or missing, then return a null list.
+     */
+    fun getComments(ids: List<Long>): Deferred<List<Comment>?> {
+        return async(ioContext) {
```
chrisbanes: Small nit: I'd use expression style here.

base/src/main/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRemoteDataSource.kt
```diff
@@ -0,0 +1,63 @@
...
+            val requestIds = ids.joinToString(",")
+            val response = service.getComments(requestIds).await()
+            if (response.isSuccessful && response.body() != null) {
+                response.body().orEmpty()
+            } else {
+                null
```
chrisbanes: No error handling?

base/src/main/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,109 @@
...
+import kotlinx.coroutines.experimental.withContext
+import kotlin.coroutines.experimental.CoroutineContext
+
+/**
+ * Repository for Designer News comments. Works with the [DesignerNewsCommentsRemoteDataSource] to
+ * get the data.
+ */
+class DesignerNewsCommentsRepository(
+    private val remoteDataSource: DesignerNewsCommentsRemoteDataSource,
+    private val uiContext: CoroutineContext = UI,
```
chrisbanes: I prefer creating a data class to hold all of the contexts, more scalable and forces you to think more generally. You don't want to be bouncing across threads accidentally.

base/src/main/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,109 @@
...
+    /**
+     * Gets comments, together will all the replies from the API. The successful result is
+     * delivered to [onSuccess]. In case of failure the error is delivered to [onError], on the
+     * [uiContext].
+     */
+    fun getComments(
+        ids: List<Long>,
+        onSuccess: (comments: List<Comment>) -> Unit,
+        onError: (error: String) -> Unit
+    ) {
```
chrisbanes: You may as well return the Job here. Plus expression style looks nicer.

base/src/main/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,109 @@
...
+        launch(uiContext) {
+            // request comments and await until the result is received.
+            val comments = getAllComments(ids)
```
chrisbanes: Probably better to run this whole method on the io thread right? Then come back to the ui context only for the `onSuccess`, etc.

base/src/main/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,109 @@
...
+
+    /**
+     * Get all comments and their replies, on the [ioContext].
+     */
+    private suspend fun getAllComments(
+        parentIds: List<Long>
+    ): List<List<Comment>>? {
+        return withContext(ioContext) {
+            val replies = mutableListOf<List<Comment>>()
+            var newReplies = remoteDataSource.getComments(parentIds).await()
```
chrisbanes: This is a bit necessary. You're switching context and then immediately waiting for another coroutine to return (if you starve threads you'll deadlock here). Personally I've been moving to a pattern when the datasources/interactors know nothing about coroutines, but the repository/VM is the one 'running' the coroutine.

base/src/main/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,109 @@
...
+            replies
+        }
+    }
+
+    private fun matchCommentsWithReplies(
+        comments: List<Comment>,
+        replies: List<Comment>
+    ): List<Comment> {
+        replies.map { reply ->
+            comments.filter { comment -> comment.id == reply.links.parentComment }
```
chrisbanes: Small nit: if you're doing a map you can inline the filter check into the map (use `mapNotNull`). One less list copy.

base/src/main/java/io/plaidapp/base/designernews/data/api/model/Comment.java
```diff
@@ -19,42 +19,50 @@
...
 import java.util.ArrayList;
 import java.util.Date;
 import java.util.List;
+import java.util.Objects;
 
 /**
  * Models a comment on a designer news story.
+ * // TODO separate the request object to the business object
  */
 public class Comment implements Parcelable {
```
chrisbanes: Can't convert this to KT? Seems like a simple data class (probably out of scope for this PR)

base/src/test/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRemoteDataSourceTest.kt
```diff
@@ -0,0 +1,99 @@
...
+            val response = dataSource.getComments(listOf(1L)).await()
+
+            // Then the response is the expected one
+            assertNotNull(response)
+            assertEquals(replies, response)
+        }
+    }
+
+    @Test
+    fun getComments_forMultipleComments() {
```
chrisbanes: Just a code style thing. I've been moving to using `runBlocking` as an [expression method](https://github.com/chrisbanes/tivi/blob/master/app/src/androidTest/java/app/tivi/interactors/TraktEpisodeSyncerTest.kt#L91).

It makes the test easier to read and it kind of hides the fact that we're testing a coroutine.

base/src/main/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,109 @@
...
+) {
+
+    /**
+     * Gets comments, together will all the replies from the API. The successful result is
+     * delivered to [onSuccess]. In case of failure the error is delivered to [onError], on the
+     * [uiContext].
+     */
+    fun getComments(
+        ids: List<Long>,
+        onSuccess: (comments: List<Comment>) -> Unit,
```
chrisbanes: Mixing coroutines and listeners feels wrong. Coroutines are designed to allow you to write imperative code which is asynchronous. This should just return a `Result` class with the result or error.

designernews/src/main/java/io/plaidapp/ui/designernews/story/DesignerNewsStory.java
```diff
@@ -579,8 +594,8 @@ private void needsLogin(View triggeringView, int requestCode) {
     private void unnestComments(List<Comment> nested, List<Comment> flat) {
         for (Comment comment : nested) {
             flat.add(comment);
-            if (comment.comments != null && comment.comments.size() > 0) {
-                unnestComments(comment.comments, flat);
+            if (comment.getReplies() != null && comment.getReplies().size() > 0) {
```
chrisbanes: Nice kotlin one statementer: `comment.getReplies()?.isNotEmpty() == true`. Would be even nicer if `getReplies()` wasn't nullable.

designernews/src/main/java/io/plaidapp/ui/designernews/story/DesignerNewsStory.java
```diff
@@ -472,7 +487,7 @@ private void bindDescription() {
         if (story.user_display_name != null && story.user_job != null) {
             SpannableString poster = new SpannableString(story.user_display_name.toLowerCase());
             poster.setSpan(new TextAppearanceSpan(this, io.plaidapp.R.style
-                    .TextAppearance_CommentAuthor),
+                            .TextAppearance_CommentAuthor),
```
chrisbanes: Does this need a new line? Doesn't seem that long

base/src/main/java/io/plaidapp/base/designernews/data/api/model/Comment.java
nickbutcher: Ah, perhaps a gson thing with how it instantiates the object?

base/src/main/java/io/plaidapp/base/designernews/data/api/model/Comment.java
florina-muntenescu: another PR. this is getting too much already

designernews/src/main/java/io/plaidapp/ui/designernews/story/DesignerNewsStory.java
florina-muntenescu: hah - true. but this is java :)

designernews/src/main/java/io/plaidapp/ui/designernews/story/DesignerNewsStory.java
florina-muntenescu: it's too long.

designernews/src/main/java/io/plaidapp/ui/designernews/story/DesignerNewsStory.java
chrisbanes: Ah, totally missed that 🤦

florina-muntenescu:
@chrisbanes  and @nickbutcher  could you take a look at this again, please? 
For now I'm keeping both coroutines and a lambda function in the Repository, because the method is called from Java. Once we add a ViewModel (and move the entire functionality in Kotlin), it will handle the coroutines

base/src/main/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
florina-muntenescu: I don't get how this applies here. the filter is on another object than the initial map.

base/src/test/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRemoteDataSourceTest.kt
florina-muntenescu: nice. i like it

base/src/main/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,129 @@
...
+                }
+            }
+        }
+        // the last request was unsuccessful
+        // if we already got some comments and replies, then use that data and ignore the error
+        return if (replies.isNotEmpty()) {
+            matchComments(replies)
+            Result.Success(replies[0])
+        } else {
+            // if we never got data, then return the error
```
chrisbanes: Nit: if we didn't get any data

base/src/main/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
chrisbanes: This is the same right?

```
comments.mapNotNull { comment ->
    if (comment.id == reply.links.parentComment) {
        comment.addReply(reply)
    } else {
        null
    }
}
```

base/src/main/java/io/plaidapp/base/designernews/data/api/model/Comment.java
```diff
@@ -19,42 +19,50 @@
...
+            String body,
+            String body_html,
+            Date created_at,
+            int depth,
+            int vote_count,
+            long user_id,
+            String user_display_name,
+            String user_portrait_url,
+            String user_job,
+            CommentLinks links,
```
chrisbanes: Run formatter on all params?

base/src/main/java/io/plaidapp/base/designernews/data/api/model/Comment.java
chrisbanes: Pssst, use Moshi.

base/src/main/java/io/plaidapp/base/designernews/data/api/model/Comment.java
```diff
@@ -130,12 +156,50 @@ public Builder setUserJob(String user_job) {
...
+                Objects.equals(user_portrait_url, comment.user_portrait_url) &&
+                Objects.equals(user_job, comment.user_job) &&
+                Objects.equals(links, comment.links) &&
+                Objects.equals(replies, comment.replies) &&
+                Objects.equals(upvoted, comment.upvoted);
+    }
+
+    @Override
+    public int hashCode() {
+
```
chrisbanes: Empty line

base/src/main/java/io/plaidapp/base/designernews/data/api/model/CommentLinks.kt
```diff
@@ -0,0 +1,37 @@
...
+import kotlinx.android.parcel.Parcelize
+
+/**
+ * Models comment links received from DesignerNews v2 API
+ */
+@Parcelize
+data class CommentLinks(
+    val user: String,
+    val story: String,
+    @SerializedName("parent_comment")
```
chrisbanes: Not a big fan of this style. If you're putting annotations above props, I'd make sure there's a line between the props. Or just put the annotations on the same line as the declaration (my preference)

base/src/main/java/io/plaidapp/base/util/CollapsingTextHelper.java
```diff
@@ -108,7 +108,7 @@ public CollapsingTextHelper(View view) {
      * A value of {@code 1.0} indicates that the layout is fully collapsed.
      */
     public void setExpansionFraction(@FloatRange(from = 0f, to = 1f) float fraction) {
-        fraction = MathUtils.constrain(fraction, 0f, 1f);
+        fraction = MathUtils.constrain( 0f, 1f, fraction);
```
chrisbanes: Nit: just use [MathUtils.clamp](https://developer.android.com/reference/android/support/v4/math/MathUtils) from X

chrisbanes: Looks good overall :rocket:

base/src/main/java/io/plaidapp/base/designernews/data/api/model/Comment.java
nickbutcher: Yup, it's in the backlog.

app/build.gradle
```diff
@@ -66,5 +66,11 @@ dependencies {
     implementation "com.crashlytics.sdk.android:crashlytics:${versions.chrashlytics}"
 }
 
+kotlin {
```
nickbutcher: Is this needed in both `:app` and `:base`?

base/src/main/java/io/plaidapp/base/data/CoroutinesContextProvider.kt
```diff
@@ -0,0 +1,26 @@
...
+package io.plaidapp.base.data
+
+import kotlinx.coroutines.experimental.CommonPool
+import kotlinx.coroutines.experimental.android.UI
+import kotlin.coroutines.experimental.CoroutineContext
+
+data class CoroutinesContextProvider(
+    val main: CoroutineContext = UI,
+    val io: CoroutineContext = CommonPool
+)
```
nickbutcher: nit: newline

base/src/main/java/io/plaidapp/base/data/api/Result.kt
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package io.plaidapp.base.data.api
+
+/**
+ * A generic class that holds a value with its loading status.
+ * @param <T>
+ */
+sealed class Result<out R> {
```
nickbutcher: Do we want to support nullabe data type `T`? i.e. should we constrain it to extend `Any`?

base/src/test/java/io/plaidapp/base/designernews/data/api/comments/DesignerNewsCommentsRepositoryTest.kt
```diff
@@ -0,0 +1,118 @@
...
+        Mockito.`when`(service.getComments("11")).thenReturn(CompletableDeferred(apiResult))
+        var result: Result<List<Comment>?>? = null
+
+        // When getting the replies
+        repository.getComments(listOf(11L)) { it -> result = it }
+
+        // Then the correct list of comments was requested from the API
+        Mockito.verify(service).getComments("11")
+        // Then the correct list is received
+        assertEquals(listOf(reply1), (result as Result.Success).data)
```
nickbutcher: Rather than looking into the result's data, should we make the expected `Response.Success(listOf(reply1))`?

base/src/test/java/io/plaidapp/base/designernews/data/api/comments/FakeAppInjection.kt
```diff
@@ -0,0 +1,23 @@
...
+ * limitations under the License.
+ */
+
+package io.plaidapp.base.designernews.data.api.comments
+
+import io.plaidapp.base.data.CoroutinesContextProvider
+import kotlinx.coroutines.experimental.Unconfined
+
+fun provideFakeCoroutinesContextProvider(): CoroutinesContextProvider =
+        CoroutinesContextProvider(Unconfined, Unconfined)
```
nickbutcher: nit: newline

base/src/main/java/io/plaidapp/base/designernews/data/api/model/Comment.java
florina-muntenescu: the code is already properly formatted, says AS

base/src/main/java/io/plaidapp/base/util/CollapsingTextHelper.java
florina-muntenescu: niiice! didn't know of that method

core/src/main/java/io/plaidapp/core/data/Result.kt
```diff
@@ -0,0 +1,41 @@
...
+ * limitations under the License.
+ */
+
+package io.plaidapp.core.data
+
+/**
+ * A generic class that holds a value with its loading status.
+ * @param <T>
+ */
+sealed class Result<out Any> {
```
nickbutcher: This type parameter doesn't seem to be used?

core/src/main/java/io/plaidapp/core/data/Result.kt
```diff
@@ -0,0 +1,41 @@
...
+
+    data class Success<out T>(val data: T) : Result<T>()
```
nickbutcher: I was [suggesting](https://github.com/nickbutcher/plaid/pull/311#discussion_r199780771) constraining the type of `data` to extend `Any` that is to not be nullable.

core/src/main/java/io/plaidapp/core/data/Result.kt
```diff
@@ -0,0 +1,41 @@
...
+            is Error -> "Error[exception=$exception]"
+            Loading -> "Loading"
+        }
+    }
+}
+
+/**
+ * `true` if [Result] is of type [Success] and holds non-null [Success.data].
+ */
+fun Result<*>.isSuccessful() = this is Result.Success && data != null
```
nickbutcher: ... which would make this extension redundant as we could just check the type.

core/src/main/java/io/plaidapp/core/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,130 @@
...
+            result
+        }
+    }
+
+    /**
+     * Build up the replies tree, by matching the replies from lower levels to the level above they
+     * belong to
+     */
+    private fun matchComments(comments: List<List<Comment>>) {
+        for (index: Int in comments.size - 1 downTo 1) {
```
nickbutcher: nit: infer type

core/src/main/java/io/plaidapp/core/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,130 @@
...
+    }
+
+    private fun matchCommentsWithReplies(
+        comments: List<Comment>,
+        replies: List<Comment>
+    ): List<Comment> {
+        // for every reply, get the comment to which the reply belongs to and add it to the list
+        // of replies for that comment
+        replies.map { reply ->
+            comments.find { comment -> comment.id == reply.links.parentComment }?.addReply(reply)
```
nickbutcher: nit: i'd find `it.id == reply...` more readable here

core/src/test/java/io/plaidapp/core/designernews/data/api/comments/DesignerNewsCommentsRepositoryTest.kt
```diff
@@ -0,0 +1,118 @@
...
+        Mockito.`when`(service.getComments("11")).thenReturn(CompletableDeferred(apiResult))
+        var result: Result<List<Comment>?>? = null
+
+        // When getting the replies
+        repository.getComments(listOf(11L)) { it -> result = it }
+
+        // Then the correct list of comments was requested from the API
+        Mockito.verify(service).getComments("11")
+        // Then the correct list is received
+        assertEquals(listOf(reply1), (result as Result.Success).data)
```
nickbutcher: Change expected to `Result.Success(listOf(reply1))` here as well.

core/src/test/java/io/plaidapp/core/designernews/data/api/comments/DesignerNewsCommentsRepositoryTest.kt
```diff
@@ -0,0 +1,118 @@
...
+        var result: Result<List<Comment>?>? = null
+
+        // When getting the comments from the repository
+        repository.getComments(listOf(1L)) { it -> result = it }
+
+        // Then  API requests were triggered
+        Mockito.verify(service).getComments("1")
+        Mockito.verify(service).getComments("11,12")
+        // Then the correct result is received
+        assertEquals(arrayListOf(parentCommentWithReplies), (result as Result.Success).data)
```
nickbutcher: ditto update expected.

core/src/test/java/io/plaidapp/core/designernews/data/api/comments/DesignerNewsCommentsRepositoryTest.kt
```diff
@@ -0,0 +1,118 @@
...
```
nickbutcher: ditto.

core/src/main/java/io/plaidapp/core/data/Result.kt
florina-muntenescu: but if we remove it, then how do we enforce generics?

nickbutcher: As agreed over chat we do need this param even if it's unused.

**But** I think we still need the constraint on the class param, that is I think `sealed class Result<out Any>` should be `sealed class Result<out R : Any>` to enforce non-null `Success.data`.

e.g. Currently `DesignerNewsCommentsRemoteDataSource.getComments` is still able to return `Result<List<Comment>?>` with this signature.

core/src/main/java/io/plaidapp/core/data/Result.kt
nickbutcher: And again I think that correctly enforcing `Success.data` to be non-null makes this extension redundant as you can just check the type.

core/src/main/java/io/plaidapp/core/designernews/data/api/comments/DesignerNewsCommentsRemoteDataSource.kt
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * Work with the Designer News API to get comments. The class knows how to construct the requests.
+ */
+class DesignerNewsCommentsRemoteDataSource(private val service: DesignerNewsService) {
+
+    /**
+     * Get a list of comments based on ids from Designer News API.
+     * If the response is not successful or missing, then return a null list.
+     */
+    suspend fun getComments(ids: List<Long>): Result<List<Comment>?> {
```
nickbutcher: Return type data shouldn't be nullable.

core/src/main/java/io/plaidapp/core/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,130 @@
...
+    /**
+     * Gets comments, together will all the replies from the API. The result is
+     * delivered to [onResult].
+     * TODO since this method is called from java, for now, let this method handle the context
+     * switching. This should be moved to a point closer to the UI, ensuring that all the
+     * business logic is away from the main context
+     */
+    fun getComments(
+        ids: List<Long>,
+        onResult: (result: Result<List<Comment>?>) -> Unit
```
nickbutcher: Result data shouldn't be nullable.

core/src/main/java/io/plaidapp/core/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,130 @@
...
+        // request comments and await until the result is received.
+        val result = withContext(contextProvider.io) { getAllComments(ids) }
+        onResult(result)
+    }
+
+    /**
+     * Get all comments and their replies. If we get an error on any reply depth level, ignore it
+     * and just use the comments retrieved until that point.
+     */
+    private suspend fun getAllComments(parentIds: List<Long>): Result<List<Comment>?> {
```
nickbutcher: Return type data shouldn't be nullable.

core/src/main/java/io/plaidapp/core/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,130 @@
...
+        val replies = mutableListOf<List<Comment>>()
+        // get the first level of comments
+        var result = remoteDataSource.getComments(parentIds)
+        // as long as we could get comments or replies to comments
+        while (result.isSuccessful()) {
```
nickbutcher: So this should become `while (result is Result.Success) {`

core/src/main/java/io/plaidapp/core/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,130 @@
...
+            val newReplies = (result as Result.Success).data
```
nickbutcher: So this should smart cast to just `val newReplies = result.data`

core/src/main/java/io/plaidapp/core/designernews/data/api/comments/DesignerNewsCommentsRemoteDataSource.kt
```diff
@@ -0,0 +1,57 @@
...
+import java.io.IOException
+
+/**
+ * Work with the Designer News API to get comments. The class knows how to construct the requests.
+ */
+class DesignerNewsCommentsRemoteDataSource(private val service: DesignerNewsService) {
+
+    /**
+     * Get a list of comments based on ids from Designer News API.
+     * If the response is not successful or missing, then return a null list.
```
nickbutcher: 'then return a null list' It no longer does this.

core/src/main/java/io/plaidapp/core/data/Result.kt
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package io.plaidapp.core.data
+
+/**
+ * A generic class that holds a value with its loading status.
+ * @param <T>
+ */
+sealed class Result<out R : Any> {
```
nickbutcher: nit: Should we make `Result` generic over `T` as well rather than `R` to communicate that it should be the same type that `Success` is generic over?

core/src/main/java/io/plaidapp/core/designernews/data/api/comments/DesignerNewsCommentsRepository.kt
```diff
@@ -0,0 +1,127 @@
...
+    }
+
+    private fun matchCommentsWithReplies(
+        comments: List<Comment>,
+        replies: List<Comment>
+    ): List<Comment> {
+        // for every reply, get the comment to which the reply belongs to and add it to the list
+        // of replies for that comment
+        replies.map { reply ->
+            comments.find { it -> it.id == reply.links.parentComment }?.addReply(reply)
```
nickbutcher: you don't need `it -> ` as it's implicit.

PR #338, opened 2018-06-25
Update Android Gradle Plugin to 3.2.0-beta02

tiembo:


nickbutcher:
Is this PR still valid? I think we already address the proguard file locations and AGP version?

tiembo:
Yep - this PR is obsolete.  Closing.

PR #342, opened 2018-06-26
Add Crashlytics.

### LINKED ISSUE #335: Add a crash reporting tool to the project

nickbutcher:
Fixes #335

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -17,6 +17,7 @@
 apply plugin: 'com.android.application'
 apply plugin: 'kotlin-android'
 apply plugin: 'kotlin-android-extensions'
+apply plugin: 'io.fabric'
 
 android {
     compileSdkVersion versions.compileSdk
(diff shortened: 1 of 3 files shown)


app/build.gradle
```diff
@@ -56,10 +57,14 @@ android {
...
 }
 
 dependencies {
     api project(':base')
     api project(':bypass')
     implementation project(':search')
+    implementation "com.crashlytics.sdk.android:crashlytics:${versions.chrashlytics}"
 }
+
+apply plugin: 'com.google.gms.google-services'
```
florina-muntenescu: should this be on top, next to the other plugins?

nickbutcher: The setup instructions explicitly say to do this. See:

https://firebase.google.com/docs/android/setup?authuser=0#manually_add_firebase
https://stackoverflow.com/a/38018985

nickbutcher: Confirmed—if I move the plugin to the top then I no longer receive crashes in the console.

PR #361, opened 2018-07-04
Cached user not updated after login

florina-muntenescu:


core/src/main/java/io/plaidapp/core/designernews/login/data/DesignerNewsLoginRepository.kt
```diff
@@ -67,10 +67,11 @@ class DesignerNewsLoginRepository(
                 { error -> onError(error) })
     }
 
-    private fun setLoggedInUser(user: User?) {
-        user?.let {
+    private fun setLoggedInUser(loggedInUser: User?) {
```
nickbutcher: Would it make sense to make this param non-null?

PR #365, opened 2018-07-04
Upvote a DN story

florina-muntenescu:


core/src/main/java/io/plaidapp/core/designernews/data/api/votes/DesignerNewsVotesRepository.kt
```diff
@@ -0,0 +1,74 @@
...
+class DesignerNewsVotesRepository(
+    private val service: DesignerNewsService,
+    private val loginRepository: DesignerNewsLoginRepository,
+    private val contextProvider: CoroutinesContextProvider
+) {
+    fun upvoteStory(
+        id: Long,
+        onResult: (result: Result<Unit>) -> Unit
+    ) = launch(contextProvider.main) {
+        val user = loginRepository.user
```
nickbutcher: Why note use `loginRepo.isLoggedId` here?

core/src/main/java/io/plaidapp/core/designernews/data/api/votes/DesignerNewsVotesRepository.kt
```diff
@@ -0,0 +1,74 @@
...
+        if (user != null) {
+            val request = UpvoteRequest(id, user.id)
+            // right now we just care whether the response is successful or not.
+            // TODO save the response in the database
+            val response = withContext(contextProvider.io) { service.upvoteStoryV2(request) }.await()
+            if (response.isSuccessful) {
+                onResult(Result.Success(Unit))
+                return@launch
+            } else {
+                onResult(Result.Error(IOException(
```
nickbutcher: We'll need to be able to differentiate the two error cases right?

PR #372, opened 2018-07-05
Set app version name in build script instead of git tag.

nickbutcher:
This is a step toward enabling releasing from CI. Version code still
derived from git commit count.


nickbutcher:
Sorry for the slightly noisey PR—in doing this I wondered what the difference between single and double quotes was in groovy. [Turns out](https://stackoverflow.com/questions/6761498/whats-the-difference-of-strings-within-single-or-double-quotes-in-groovy) single quotes are for regular strings, double quotes are for templated strings. I then had to fix it throughout.

PR #354, opened 2018-06-29
Move proguard files to more suitable locations

keyboardsurfer:

PR #376, opened 2018-07-06
Proguard cleanup

### LINKED ISSUE #363: Proguard setup issues
Address @wojtek-kalicinski's comments on proguard changes.

https://github.com/nickbutcher/plaid/pull/354#pullrequestreview-134016711

keyboardsurfer:
Fix #363
Address dangling comments of #354


about/build.gradle
```diff
@@ -32,9 +32,7 @@ android {
     }
 
     buildTypes {
-        release {
-            proguardFiles 'proguard-rules.pro'
-        }
+        release { }
```
nickbutcher: Can we remove this empty block?

app/build.gradle
```diff
@@ -41,7 +41,8 @@ android {
     buildTypes {
         release {
             minifyEnabled true
-            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'), 'proguard-rules.pro'
+            proguardFiles getDefaultProguardFile('proguard-android-optimize.txt'),
```
nickbutcher: Why wrap here?

about/build.gradle
nickbutcher: Same throughout.

app/build.gradle
keyboardsurfer: Line length

PR #393, opened 2018-07-12
Update gradle tool to 3.3.0-alpha03

XinyueZ:


nickbutcher:
Thanks but we got this in #413

PR #394, opened 2018-07-13
Moving CommentViewHolder binding logic away from the Activity

florina-muntenescu:


designernews/src/main/java/io/plaidapp/ui/designernews/story/CommentViewHolder.kt
```diff
@@ -130,4 +149,18 @@ internal class CommentViewHolder(itemView: View) : RecyclerView.ViewHolder(itemV
...
         }
     }
+
+    fun setExpanded(expanded: Boolean) {
+        if (expanded) {
+            val threadDepthWidth = threadDepth.getDrawable().getIntrinsicWidth()
+            val leftShift = (-(threadDepthWidth + (threadDepth.getLayoutParams() as ViewGroup.MarginLayoutParams).marginEnd)).toFloat()
+            author.translationX = leftShift
+            comment.translationX = leftShift
+            threadDepth.translationX = (-(threadDepthWidth + (threadDepth.getLayoutParams() as ViewGroup.MarginLayoutParams).marginStart)).toFloat()
```
keyboardsurfer: Can be casted once by holding on to a `val`.
Would make this calculation easier to read.

designernews/src/main/java/io/plaidapp/ui/designernews/story/CommentViewHolder.kt
```diff
@@ -28,13 +28,32 @@ import io.plaidapp.core.util.AnimUtils.getFastOutSlowInInterpolator
...
+    private val threadGap: Int
+) : RecyclerView.ViewHolder(itemView) {
+
+    private val threadDepth: ImageView = itemView.findViewById(R.id.depth)
+    private val author: AuthorTextView = itemView.findViewById(R.id.comment_author)
+    private val timeAgo: TextView = itemView.findViewById(R.id.comment_time_ago)
+    val comment: TextView = itemView.findViewById(R.id.comment_text)
 
-internal class CommentViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
+    init {
```
keyboardsurfer: Why not use `apply` during initialization above?

designernews/src/main/java/io/plaidapp/ui/designernews/story/DesignerNewsStory.java
```diff
@@ -527,6 +521,25 @@ private void bindDescription() {
         }
     }
 
+    private CharSequence getStoryPosterTimeText(
```
keyboardsurfer: We're still in Java country here.

designernews/src/main/java/io/plaidapp/ui/designernews/story/CommentViewHolder.kt
florina-muntenescu: doesn't work

PR #400, opened 2018-07-16
Removing permission request from DN Login

florina-muntenescu:


designernews/src/main/java/io/plaidapp/ui/designernews/DesignerNewsLogin.java
```diff
@@ -141,14 +118,6 @@ private void bindViews() {
...
     public void onEnterAnimationComplete() {
-        /* Postpone some of the setup steps so that we can run it after the enter transition (if
-        there is one). Otherwise we may show the permissions dialog or account dropdown during the
-        enter animation which is jarring. */
-        if (shouldPromptForPermission) {
-            requestPermissions(new String[]{Manifest.permission.GET_ACCOUNTS},
-                    PERMISSIONS_REQUEST_GET_ACCOUNTS);
-            shouldPromptForPermission = false;
-        }
         username.setOnFocusChangeListener((v, hasFocus) -> maybeShowAccounts());
```
nickbutcher: Both the focus listener and the `maybeShowAccounts` method can be removed as well I think.

PR #408, opened 2018-07-18
Introduce Markdown and extract LoadImageCallback

keyboardsurfer:


third_party/bypass/src/main/java/in/uncod/android/bypass/LoadImageCallback.kt
```diff
@@ -0,0 +1,12 @@
...
+import `in`.uncod.android.bypass.style.ImageLoadingSpan
+
+interface LoadImageCallback {
+    /**
+     * A callback to load an image found in a markdown document.
+     * @param src The source (url) of the image.
+     * @param loadingSpan A placeholder span making where the image should be inserted.
+     */
+    fun loadImage(src: String, loadingSpan: ImageLoadingSpan)
+}
```
florina-muntenescu: new line

third_party/bypass/src/main/java/in/uncod/android/bypass/LoadImageCallback.kt
```diff
@@ -0,0 +1,12 @@
+package `in`.uncod.android.bypass
+
+import `in`.uncod.android.bypass.style.ImageLoadingSpan
+
+interface LoadImageCallback {
```
florina-muntenescu: doc the callback

third_party/bypass/src/main/java/in/uncod/android/bypass/LoadImageCallback.kt
```diff
@@ -0,0 +1,12 @@
...
+    /**
+     * A callback to load an image found in a markdown document.
```
florina-muntenescu: this should be rather: `Callback called when an image found in a markdown document should be loaded`

PR #410, opened 2018-07-18
Convert DribbbleShot to kotlin.

### LINKED ISSUE #381: [DR] Convert DribbbleShot to Kotlin

nickbutcher:
Fixes #381


dribbble/src/main/java/io/plaidapp/dribbble/ui/ShareDribbbleImageTask.java
```diff
@@ -48,11 +48,12 @@
     @Override
     protected File doInBackground(Void... params) {
         final String url = shot.getImages().best();
+        final int[] size = shot.getImages().bestSize();
         try {
             return Glide
                     .with(activity)
                     .load(url)
-                    .downloadOnly(shot.getWidth(), shot.getHeight())
+                    .downloadOnly(size[0], size[1])
```
florina-muntenescu: the array isn't guaranteed to have 2 elements by default. could you add a check for array size?

nickbutcher: Converted sizes to an enum to be moar type safe. Good catch!

dribbble/src/main/java/io/plaidapp/dribbble/ui/ShareDribbbleImageTask.java
```diff
@@ -48,11 +48,12 @@
     @Override
     protected File doInBackground(Void... params) {
         final String url = shot.getImages().best();
+        final int[] size = shot.getImages().bestSize();
         try {
             return Glide
```
XinyueZ: Changes request here using new `Glide` API
```
                 Glide
                    .with(activity)
                    .asFile()
                    .load(url)
                    .submit(width, height)
                    .get();
```

dribbble/src/main/java/io/plaidapp/dribbble/ui/ShareDribbbleImageTask.java
XinyueZ: @florina-muntenescu `DribbbleSearchConverter.parseShot()` will never consturct `Shot` with  `width` and `height`.  Is that really the commitment?

nickbutcher:
> [will never construct `Shot` with `width` and `height`.](https://github.com/nickbutcher/plaid/pull/410#discussion_r203336272)

Right this was a legacy of the v1 API, i've removed the `width` and `height` properties from `Shot` and we now just derive the image size based on whether the url for a high quality image is present.

> [Changes request here using new Glide API](https://github.com/nickbutcher/plaid/pull/410#pullrequestreview-138201534)

I was trying to limit the scope of this PR, would you like to prepare this change?

XinyueZ:
@nickbutcher ok, I have #411 just now.

PR #423, opened 2018-07-23
Fix null-exp warning at ImeUtils.hideIme()

XinyueZ:


keyboardsurfer:
Thanks for your continued contributions. These files will be converted to Kotlin during this project. So we currently prefer not to touch them until they are translated.

PR #424, opened 2018-07-23
Fix warnings of IDE/Lint about null check before "instanceof"

XinyueZ:
I am not sure whether "null" check is the commitment of the project or not before `instanceof` being used, however, the IDE make warnings really. 
If it is commitment, this PR can be ignored.


keyboardsurfer:
Thanks for your continued contributions. These files will be converted to Kotlin during this project. So we currently prefer not to touch them until they are translated.

PR #432, opened 2018-07-24
Fix lint/IDE warnings in PocketUtils

XinyueZ:


florina-muntenescu:
As stated in the [Readme](https://github.com/nickbutcher/plaid) we're currently not accepting PRs. 
We're in progress of converting files to Kotlin so contributions to Java files are not required.

PR #437, opened 2018-07-25
Removes Dribbble prefixes, no logic changes.

nickbutcher:
Also corrected package structure in `:core` `androidTest`.


dribbble/src/main/java/io/plaidapp/dribbble/ui/shot/ShotViewModel.kt
```diff
@@ -30,11 +30,11 @@ import kotlinx.coroutines.experimental.Job
...
 
 /**
- * View model for [DribbbleShotActivity].
+ * View model for [ShotActivity].
  */
-class DribbbleShotViewModel(
+class ShotViewModel(
     shotId: Long,
-    shotRepository: DribbbleRepository,
+    shotRepository: ShotsRepository,
```
florina-muntenescu: `shotsRepository` (plural)

nickbutcher: done.

PR #440, opened 2018-07-27
Convert AuthInterceptor to Kotlin

tiembo:


core/src/main/java/io/plaidapp/core/data/api/AuthInterceptor.kt
```diff
@@ -0,0 +1,33 @@
+/*
```
florina-muntenescu: This is only used by product hunt. could you also move it to the right package?

tiembo: Done

PR #455, opened 2018-07-31
Skip rest of build if running on a forked repo

tiembo:


scripts/ftl_setup.sh
```diff
@@ -2,9 +2,10 @@
 
 # Check if this script is running on the main, non-forked repository.
 # Environment variables are not set in forked repository builds;
-# skip Firebase Test Lab steps in this case.
+# in this case, skip Firebase Test Lab steps and finish the build.
 if [ -z "$IS_MAIN_PLAID_REPO" ]; then
   echo "Running build on a forked repository - skipping FTL tests."
+  circleci step halt
```
ataulm: should this script be included in this repo, or part of the job itself? A third party patch could edit this file so that their PR does run on FTL (or worse).

tiembo: That's a good concern to bring up.  Nothing in this or the other Continuous Integration scripts are private except for the $GCLOUD_SERVICE_KEY.  This and all environment variables are set on the CI instance and are not exposed publicly.

A user can actually use their Firebase / Google cloud test key and run these scripts to test Plaid against their own Firebase Test Lab account.

PR #461, opened 2018-08-01
about 🗡

keyboardsurfer:
Introduce first dagger component and modules

AboutComponent holds modules used in about feature.
Modules for use only within about feature as well
as a module for Markdown which is declared in core
but can be used in components created within other features.

### DIFF
--- a/about/src/main/java/io/plaidapp/about/dagger/AboutActivityModule.kt
+++ b/about/src/main/java/io/plaidapp/about/dagger/AboutActivityModule.kt
@@ -38,5 +38,5 @@ import javax.inject.Singleton
 
     @Provides
     @Singleton
-    fun provideAboutStyler(activity: AboutActivity) = AboutStyler(activity)
+    fun provideAboutStyler() = AboutStyler(activity)
 }
--- a/about/src/main/java/io/plaidapp/about/dagger/AboutActivityModule.kt
+++ b/about/src/main/java/io/plaidapp/about/dagger/AboutActivityModule.kt
@@ -1,0 +1,42 @@
+/*
+ * Copyright 2018 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.plaidapp.about.dagger
+
+import android.content.res.Resources
+import dagger.Module
+import dagger.Provides
+import io.plaidapp.about.ui.AboutActivity
+import io.plaidapp.about.ui.AboutStyler
+import javax.inject.Singleton
+
+/**
+ * Dagger module providing stuff from [AboutActivity].
+ */
+@Module class AboutActivityModule(private val activity: AboutActivity) {
+
+    @Provides
+    @Singleton
+    fun provideContext(): AboutActivity = activity
+
+    @Provides
+    @Singleton
+    fun provideResources(): Resources = activity.resources
+
+    @Provides
+    @Singleton
+    fun provideAboutStyler(activity: AboutActivity) = AboutStyler(activity)
+}
--- a/about/src/main/java/io/plaidapp/about/ui/AboutActivity.kt
+++ b/about/src/main/java/io/plaidapp/about/ui/AboutActivity.kt
@@ -25,6 +25,7 @@ import android.support.v7.app.AppCompatActivity
 import android.transition.TransitionInflater
 import androidx.core.net.toUri
 import io.plaidapp.about.R
+import io.plaidapp.about.dagger.inject
 import io.plaidapp.about.ui.adapter.AboutPagerAdapter
 import io.plaidapp.about.ui.model.AboutViewModel
 import io.plaidapp.about.ui.model.AboutViewModelFactory
@@ -32,6 +33,7 @@ import io.plaidapp.about.ui.widget.InkPageIndicator
 import io.plaidapp.core.ui.widget.ElasticDragDismissFrameLayout
 import io.plaidapp.core.util.customtabs.CustomTabActivityHelper
 import io.plaidapp.core.util.event.EventObserver
+import javax.inject.Inject
 import io.plaidapp.R as appR
 
 /**
@@ -42,6 +44,8 @@ import io.plaidapp.R as appR
  */
 class AboutActivity : AppCompatActivity() {
 
+    @Inject internal lateinit var aboutViewModelFactory: AboutViewModelFactory
+
     override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_about)
@@ -50,7 +54,8 @@ class AboutActivity : AppCompatActivity() {
         val pager = findViewById<ViewPager>(R.id.pager)
         val pageIndicator = findViewById<InkPageIndicator>(R.id.indicator)
 
-        val aboutViewModelFactory = AboutViewModelFactory(AboutStyler(this), resources)
+        inject()
+
         val viewModel = ViewModelProviders
             .of(this, aboutViewModelFactory)
             .get(AboutViewModel::class.java)
--- a/about/src/main/java/io/plaidapp/about/ui/AboutStyler.kt
+++ b/about/src/main/java/io/plaidapp/about/ui/AboutStyler.kt
@@ -16,18 +16,23 @@
 
 package io.plaidapp.about.ui
 
+import android.content.res.ColorStateList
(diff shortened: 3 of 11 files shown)


about/src/main/java/io/plaidapp/about/ui/AboutStyler.kt
```diff
@@ -21,11 +21,12 @@ import android.support.annotation.ColorInt
...
 import io.plaidapp.core.util.ColorUtils
+import javax.inject.Inject
 
 /**
  * Provide style colors to links and highlights in about activity.
  */
-class AboutStyler(activity: Activity) {
+class AboutStyler @Inject constructor(activity: Activity) {
 
     val linksColor = ContextCompat.getColorStateList(activity, coreR.color.plaid_links)!!
```
dpreussler: i guess you want the activity for getting a themed context? 
As normally I would avoid injecting activities, it opens pandoras box. You could use application context that you always have in the modules.

Or maybe pass the Activitiy when needed. A bit tricky here but you could do

val Activity.linksColor and then do sth like:
```
with(appstyler) {
 ... = activity.linksColor
}
```

just as an idea

about/src/main/java/io/plaidapp/about/ui/AboutActivity.kt
```diff
@@ -50,7 +58,14 @@ class AboutActivity : AppCompatActivity() {
...
         val pageIndicator = findViewById<InkPageIndicator>(R.id.indicator)
 
-        val aboutViewModelFactory = AboutViewModelFactory(AboutStyler(this), resources)
+        DaggerAboutComponent.builder()
+            .aboutActivityModule(AboutActivityModule(this))
+            .aboutStylerModule(AboutStylerModule())
+            .markdownModule(MarkdownModule(resources.displayMetrics))
+            .build().apply {
+                inject(this@AboutActivity)
+            }
```
dpreussler: normally i try to hide this somewhere, so have some encapsulation around Dagger

florina-muntenescu:
LGTM. Your choice if you want to delay the merging this to try another DI/SL

about/src/main/java/io/plaidapp/about/ui/model/AboutViewModelFactory.kt
```diff
@@ -16,28 +16,25 @@
...
+import javax.inject.Inject
 
 /**
  * Factory to create [AboutViewModel]
  */
-internal class AboutViewModelFactory(
-    private val aboutStyler: AboutStyler,
+
+internal class AboutViewModelFactory @Inject constructor(
     val resources: Resources
```
XinyueZ: Can this `resources` be `private`?
Who will use the `resources`?   😕

about/src/main/java/io/plaidapp/about/ui/AboutStyler.kt
```diff
@@ -16,16 +16,16 @@
...
+import javax.inject.Inject
 
 /**
- * Provide style colors to links and highlights in about activity.
+ * Provide style colors to links and highlights in [AboutActivity].
  */
-class AboutStyler(activity: Activity) {
+class AboutStyler @Inject constructor(activity: AboutActivity) {
 
     val linksColor = ContextCompat.getColorStateList(activity, coreR.color.plaid_links)!!
```
XinyueZ: The `AppCompatResources.getColorStateList` is better, you can save `!!` .

about/src/main/java/io/plaidapp/about/dagger/AboutActivityModule.kt
```diff
@@ -0,0 +1,42 @@
...
+import io.plaidapp.about.ui.AboutStyler
+import javax.inject.Singleton
+
+/**
+ * Dagger module providing stuff from [AboutActivity].
+ */
+@Module class AboutActivityModule(private val activity: AboutActivity) {
+
+    @Provides
+    @Singleton
```
hzsweers: FWIW: Conventionally, `@Singleton` tends to be synonymous with something that's app-scope, rather than activity-scoped. In this case you're sort of fine since this is the root, but if an application scope is introduced later as a parent of this you'll likely want to switch to something like `@PerActivity` to avoid confusion/accidentally leaking.

about/src/main/java/io/plaidapp/about/dagger/AboutActivityModule.kt
```diff
@@ -0,0 +1,42 @@
...
+import dagger.Module
+import dagger.Provides
+import io.plaidapp.about.ui.AboutActivity
+import io.plaidapp.about.ui.AboutStyler
+import javax.inject.Singleton
+
+/**
+ * Dagger module providing stuff from [AboutActivity].
+ */
+@Module class AboutActivityModule(private val activity: AboutActivity) {
```
hzsweers: You could make this module stateless by accepting this parameter in your component instead via `@BindsInstance`. Same for the `MarkdownModule` farther down, which could accept no params if you expose `DisplayMetrics` from this module as part of your resources provisioning.

```kotlin
interface AboutComponent {

  fun inject(activity: AboutActivity)

  @Component.Builder
  interface Builder {
    fun build(): AboutComponent
  
    @BindsInstance fun activity(activity: AboutActivity): Builder
  }
}
```

Then your injection could just look like this:

```kotlin
fun AboutActivity.inject() {
    DaggerAboutComponent.builder()
        .activity(this)
        .build()
        .inject(this)
}
```

With it stateless, you could technically also make the below functions `@JvmStatic`, but that ends up looking weird in Kotlin and maybe not worth the runtime allocation savings since it wouldn't need to instantiate the module.

about/src/main/java/io/plaidapp/about/ui/AboutActivity.kt
```diff
@@ -50,7 +54,8 @@ class AboutActivity : AppCompatActivity() {
         val pager = findViewById<ViewPager>(R.id.pager)
```
XinyueZ: The project is `databinding` enabled, why not use it here in order to reduce performance shadow of `findViewById` 🤒 ?

about/src/main/java/io/plaidapp/about/ui/model/AboutViewModelFactory.kt
```diff
@@ -16,28 +16,25 @@
...
-import io.plaidapp.about.ui.AboutStyler
+import javax.inject.Inject
 
 /**
  * Factory to create [AboutViewModel]
  */
-internal class AboutViewModelFactory(
-    private val aboutStyler: AboutStyler,
+
+internal class AboutViewModelFactory @Inject constructor(
```
ashdavies: What would be your thoughts on using an approach with Dagger Multibindings? Where you could create a single factory for all injections. I think it probably works out to the same amount of boilerplate, but you end up not having to repeat the constructor parameters with the injection parameters.

Eg.
https://gist.github.com/ashdavies/2a96facbbe766f404d1b770182255da9

about/src/main/java/io/plaidapp/about/dagger/AboutActivityModule.kt
keyboardsurfer: That makes a lot of sense if we get to use dagger-android.
In my current understanding we can't do that since a Subcomponent needs to be registered with the AppComponent's AndroidInjector. And in our case the app module doesn't know about the dynamic-feature modules.

about/src/main/java/io/plaidapp/about/dagger/AboutActivityModule.kt
keyboardsurfer: 👍

about/src/main/java/io/plaidapp/about/ui/AboutActivity.kt
keyboardsurfer: databinding is enabled in some features but not currently in this one. I don't see much benefit from enabling at this point for `about`.

about/src/main/java/io/plaidapp/about/ui/model/AboutViewModelFactory.kt
keyboardsurfer: Due to modularity of the app I don't see how we can make use of dagger-android and Subcomponents due to the dependency relationship between app and feature modules. From what I understand, what you're suggesting relies on benefits from that relationship.

ashdavies: AFAIK it doesn't make use of `dagger-android` just creating a factory for the application view models, I'm still looking for an ideal solution, but I was looking at this implementation again, and thought that since you're injecting the `ViewModelProvider.Factory` with an injected instance of the view model, doesn't that result in a view model being instantiated regardless of whether it's used inside the activity?

keyboardsurfer: This ViewModel is being used in the activity every single time.
For ViewModels that are not being used every time we probably could do something else.

ashdavies: Yep, of course, I mean that an instance of `AboutViewModelFactory`, is being injected in `AboutActivity`, and is being supplied to the `ViewModelProviders` to ensure that the persisted view model is being used. 

My question is that since the factory has an injected `private val` rather than a JSR330 `Provider` (for example) which means the factory will always be instantiated with a *new* view model, despite the fact that the persisted view model will instead be used. 

Unless I'm reading it incorrectly?

PR #467, opened 2018-08-05
Adds try/catch for all network requests

### LINKED ISSUE #466: App crashes when doing any network requests on airplane mode

florina-muntenescu:
Fixes #466

### DIFF
--- a/core/src/main/java/io/plaidapp/core/designernews/data/comments/CommentsRemoteDataSource.kt
+++ b/core/src/main/java/io/plaidapp/core/designernews/data/comments/CommentsRemoteDataSource.kt
@@ -19,6 +19,7 @@ package io.plaidapp.core.designernews.data.comments
 import io.plaidapp.core.data.Result
 import io.plaidapp.core.designernews.data.api.DesignerNewsService
 import io.plaidapp.core.designernews.data.comments.model.CommentResponse
+import io.plaidapp.core.util.safeApiCall
 import java.io.IOException
 
 /**
@@ -29,7 +30,12 @@ class CommentsRemoteDataSource(private val service: DesignerNewsService) {
     /**
      * Get a list of comments based on ids from Designer News API.
      */
-    suspend fun getComments(ids: List<Long>): Result<List<CommentResponse>> {
+    suspend fun getComments(ids: List<Long>) = safeApiCall(
+        call = { requestGetComments(ids) },
+        errorMessage = "Error getting comments"
+    )
+
+    private suspend fun requestGetComments(ids: List<Long>): Result<List<CommentResponse>> {
         val requestIds = ids.joinToString(",")
         val response = service.getComments(requestIds).await()
         if (response.isSuccessful) {
@@ -38,7 +44,9 @@ class CommentsRemoteDataSource(private val service: DesignerNewsService) {
                 return Result.Success(body)
             }
         }
-        return Result.Error(IOException("Error getting comments ${response.code()} ${response.message()}"))
+        return Result.Error(
+            IOException("Error getting comments ${response.code()} ${response.message()}")
+        )
     }
 
     companion object {
--- a/core/src/main/java/io/plaidapp/core/util/NetworkUtils.kt
+++ b/core/src/main/java/io/plaidapp/core/util/NetworkUtils.kt
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2018 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.plaidapp.core.util
+
+import io.plaidapp.core.data.Result
+import java.io.IOException
+
+/**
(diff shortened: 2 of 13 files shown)


core/src/main/java/io/plaidapp/core/designernews/data/comments/CommentsRemoteDataSource.kt
```diff
@@ -31,18 +31,23 @@ class CommentsRemoteDataSource(private val service: DesignerNewsService) {
...
+            val response = service.getComments(requestIds).await()
+            if (response.isSuccessful) {
+                val body = response.body()
+                if (body != null) {
+                    return Result.Success(body)
+                }
             }
+            return Result.Error(IOException("Error getting comments ${response.code()} ${response.message()}"))
+        } catch (e: Exception) {
+            return Result.Error(IOException("Error getting comments  ${e.message}"))
```
nickbutcher: Let's pass `e` as the cause to the `IOException` constructor.

Same throughout.

core/src/main/java/io/plaidapp/core/designernews/data/comments/CommentsRemoteDataSource.kt
```diff
@@ -31,18 +31,23 @@ class CommentsRemoteDataSource(private val service: DesignerNewsService) {
...
         }
-        return Result.Error(IOException("Error getting comments ${response.code()} ${response.message()}"))
     }
 
     companion object {
-        @Volatile private var INSTANCE: CommentsRemoteDataSource? = null
+        @Volatile
```
nickbutcher: revert.

core/src/main/java/io/plaidapp/core/designernews/data/comments/CommentsRemoteDataSource.kt
```diff
@@ -31,18 +31,23 @@ class CommentsRemoteDataSource(private val service: DesignerNewsService) {
      */
     suspend fun getComments(ids: List<Long>): Result<List<CommentResponse>> {
         val requestIds = ids.joinToString(",")
-        val response = service.getComments(requestIds).await()
-        if (response.isSuccessful) {
-            val body = response.body()
-            if (body != null) {
-                return Result.Success(body)
+        try {
```
keyboardsurfer: Why not introduce a function that takes the original API call incl error as function param and perform the try/catch in there, wrapping a potential exception in a `Result.Error`?

Something along the lines of this:
```
fun safeApiCall(message: String, call: () -> Result) {
    return try {
        call.invoke()
    } catch (e: Exception) {
        Result.Error(IOException("Error $message", e)
    }
}
```

This way the error handling code would be centralized.
The call site could then look like this:

```
safeApiCall("getting comments", getComments())
```

core/src/main/java/io/plaidapp/core/util/NetworkUtils.kt
```diff
@@ -0,0 +1,28 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.plaidapp.core.util
+
+import io.plaidapp.core.data.Result
+import java.io.IOException
+
+suspend fun <T : Any> safeApiCall(message: String, call: suspend () -> Result<T>): Result<T> {
```
keyboardsurfer: Document

core/src/main/java/io/plaidapp/core/util/NetworkUtils.kt
```diff
@@ -0,0 +1,32 @@
...
+package io.plaidapp.core.util
+
+import io.plaidapp.core.data.Result
+import java.io.IOException
+
+/**
+ * Wrap a suspending API [call] in try/catch. In case an exception is thrown, a [Result.Error] is
+ * created based on the [message].
+ */
+suspend fun <T : Any> safeApiCall(message: String, call: suspend () -> Result<T>): Result<T> {
```
nickbutcher: Can this be marked `inline`?

core/src/main/java/io/plaidapp/core/util/NetworkUtils.kt
```diff
@@ -0,0 +1,32 @@
...
+    return try {
+        call.invoke()
```
nickbutcher: Just `call()` would be cleaner.

core/src/main/java/io/plaidapp/core/util/NetworkUtils.kt
nickbutcher: Can we rename `message` to `errorMessage` to make its purpose clearer?

core/src/main/java/io/plaidapp/core/designernews/data/comments/CommentsRemoteDataSource.kt
```diff
@@ -29,7 +30,12 @@ class CommentsRemoteDataSource(private val service: DesignerNewsService) {
     /**
      * Get a list of comments based on ids from Designer News API.
      */
-    suspend fun getComments(ids: List<Long>): Result<List<CommentResponse>> {
+    suspend fun getComments(ids: List<Long>) = safeApiCall(
+        message = "Error getting comments",
+        call = { requestGetComments(ids) }
```
nickbutcher: I think that we should either move this outside of the parens **or** make `call` the first param as that is the expected case.

nickbutcher: Same throughout.

core/src/main/java/io/plaidapp/core/util/NetworkUtils.kt
```diff
@@ -0,0 +1,32 @@
...
+    } catch (e: Exception) {
```
nickbutcher: Perhaps comment why we have to catch `Exception` here.

core/src/main/java/io/plaidapp/core/util/NetworkUtils.kt
florina-muntenescu: it can't be inline because we're calling a suspend function

PR #469, opened 2018-08-05
Update gradle to 3.3.0-alpha05

XinyueZ:
Update gradle to 3.3.0-alpha04


XinyueZ:
I am very curious, this PR actually appeared CI error, is 3.3.0-alpha04 not compatible with some bugs?

PR #472, opened 2018-08-06
LoggedInUser dao changes - delete and set user

tiembo:


core/src/main/java/io/plaidapp/core/data/LoggedInUserDao.kt
```diff
@@ -21,16 +21,26 @@ import android.arch.persistence.room.Dao
 import android.arch.persistence.room.Insert
```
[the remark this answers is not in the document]
tiembo: Moved.

core/src/main/java/io/plaidapp/core/designernews/data/LoggedInUserDao.kt
```diff
@@ -14,23 +14,38 @@
...
 import android.arch.persistence.room.Query
+import android.arch.persistence.room.Transaction
 import io.plaidapp.core.designernews.data.users.model.LoggedInUser
 
 /**
  * This Data Access Object handles Room database operations for the [LoggedInUser] class.
  */
 @Dao
-interface LoggedInUserDao {
+abstract class LoggedInUserDao {
```
Sripadmanabans: I apologise if I am not allowed to comment. I am just curious as to the reason why we could not use default interface here?

florina-muntenescu: because `setLoggedInUser` is a concrete method

core/src/main/java/io/plaidapp/core/designernews/data/LoggedInUserDao.kt
```diff
@@ -14,23 +14,38 @@
...
     @Query("SELECT * FROM logged_in_user LIMIT 1")
-    fun getLoggedInUser(): LiveData<LoggedInUser>
+    abstract fun getLoggedInUser(): LiveData<LoggedInUser>
 
+    @Transaction
+    open fun setLoggedInUser(loggedInUser: LoggedInUser) {
```
florina-muntenescu: add a comment describing what this method does

core/src/main/java/io/plaidapp/core/designernews/data/LoggedInUserDao.kt
Sripadmanabans: Since this class does not involve state or constructors and ```setLoggerInUser()``` is open isn't is just easier to write it as a function on the interface itself? I was wondering if using a "Default Interface" has an impact in performance.

core/src/main/java/io/plaidapp/core/designernews/data/LoggedInUserDao.kt
tiembo: Added.

PR #474, opened 2018-08-13
[DN] Moves comments triggering from activity to view model

florina-muntenescu:


designernews/src/main/java/io/plaidapp/designernews/domain/CommentsUseCase.kt
```diff
@@ -14,37 +14,27 @@
...
-    ) = launch(contextProvider.io) {
+    suspend operator fun invoke(ids: List<Long>): Result<List<Comment>> {
         // Get the comments with replies
         val commentsWithRepliesResult = commentsWithRepliesUseCase.getCommentsWithReplies(ids)
         if (commentsWithRepliesResult is Result.Error) {
-            withContext(contextProvider.main) {
-                onResult(Result.Error(commentsWithRepliesResult.exception))
-            }
-            return@launch
+            return Result.Error(commentsWithRepliesResult.exception)
```
nickbutcher: Why re-wrap it rather than just returning result directly?

designernews/src/main/java/io/plaidapp/designernews/ui/story/StoryActivity.java
```diff
@@ -146,21 +143,12 @@ protected void onCreate(Bundle savedInstanceState) {
...
-        commentsUseCase.getComments(story.getLinks().getComments(),
-                result -> {
-                    if (result instanceof Result.Success) {
-                        Result.Success<List<Comment>> success = (Result.Success<List<Comment>>) result;
-                        List<Comment> data = success.getData();
-                        setupComments(data);
-                    }
-                    return Unit.INSTANCE;
-                });
+        viewModel.getUiState().observe(this,
```
nickbutcher: Conflicting naming. Is it a state or model?

designernews/src/main/java/io/plaidapp/designernews/ui/story/StoryViewModel.kt
```diff
@@ -34,20 +38,26 @@ import kotlinx.coroutines.experimental.withContext
 class StoryViewModel(
     storyId: Long,
     getStoryUseCase: GetStoryUseCase,
+    private val commentsUseCase: CommentsUseCase,
```
nickbutcher: Didn't we agree to drop the `UseCase` suffix for instances? This also reveals that `CommentsUseCase` isn't super descriptive; what is it doing with comments?

designernews/src/main/java/io/plaidapp/designernews/domain/CommentsUseCase.kt
florina-muntenescu: done

designernews/src/main/java/io/plaidapp/designernews/ui/story/StoryActivity.java
florina-muntenescu: done

designernews/src/main/java/io/plaidapp/designernews/ui/story/StoryViewModel.kt
florina-muntenescu: forgot about that. done now.

designernews/src/main/java/io/plaidapp/designernews/ui/story/StoryViewModel.kt
```diff
@@ -73,4 +83,26 @@ class StoryViewModel(
         parentJob.cancel()
         super.onCleared()
     }
+
+    private fun getComments() = launch(contextProvider.io, parent = parentJob) {
+        story.links?.let {
+            val result = commentsWithRepliesAndUsers(it.comments)
+            if (result is Result.Success) {
```
tiembo: Optionally add comment here that Result.Error is ignored

PR #476, opened 2018-08-14
Add upvotes property to LoggedInUser

tiembo:
Includes type converter for storing/retrieving upvotes from database.

### DIFF
--- a/core/src/main/java/io/plaidapp/core/data/Converters.kt
+++ b/core/src/main/java/io/plaidapp/core/data/Converters.kt
@@ -1,0 +1,38 @@
+/*
+ * Copyright 2018 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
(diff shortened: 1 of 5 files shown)


core/src/main/java/io/plaidapp/core/data/Converters.kt
```diff
@@ -0,0 +1,38 @@
...
+
+package io.plaidapp.core.data
+
+import android.arch.persistence.room.TypeConverter
+
+/**
+ * Type converters to allow Room to reference complex data types.
+ */
+class Converters {
+    private val delimiter = ","
```
nickbutcher: `const`

core/src/main/java/io/plaidapp/core/designernews/data/users/model/LoggedInUser.kt
```diff
@@ -42,5 +42,9 @@ data class LoggedInUser(
 
     @ColumnInfo(name = "potrait_url")
     @SerializedName("portrait_url")
-    val portraitUrl: String? = null
+    val portraitUrl: String? = null,
+
+    @ColumnInfo(name = "upvotes")
+    @SerializedName("upvotes")
+    val upvotes: List<String>
```
nickbutcher: Elsewhere we're converting IDs to `Long`s. Should we also do this here?

florina-muntenescu: +1 for `Long`

core/src/main/java/io/plaidapp/core/data/Converters.kt
tiembo: Done

core/src/main/java/io/plaidapp/core/designernews/data/users/model/LoggedInUser.kt
tiembo: Done

PR #479, opened 2018-08-18
Optimize flattening CommentWithReplies

hzsweers:
This tweaks flattening `CommentWithReplies` with a more optimized version via recursive sequence-generating `flattenWithReplies` property. This makes the entire flattening process a single iteration with no mutable intermediate state. 

Also opportunistically changed

```kotlin
val userMapping = users.map { it.id to it }.toMap()
```

to the more idiomatic (but functionally the same sans one iteration)

```kotlin
val userMapping = users.associateBy(User::id)
```

### DIFF
--- a/core/src/main/java/io/plaidapp/core/designernews/domain/model/CommentWithReplies.kt
+++ b/core/src/main/java/io/plaidapp/core/designernews/domain/model/CommentWithReplies.kt
@@ -17,7 +17,7 @@
 package io.plaidapp.core.designernews.domain.model
 
 import io.plaidapp.core.designernews.data.users.model.User
-import java.util.Date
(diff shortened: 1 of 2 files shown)


hzsweers:
The build failure looks like a flake/unrelated? That test passes for me locally 🤔

tiembo:
Hey @hzsweers - agree that this PR shouldn't be failing the build.  I manually re-ran it but it still failed.  Mind rebasing and we'll give it another try?

hzsweers:
Done!

tiembo:
Thanks! Looks like there's an issue with that test which we'll look into. (cc @XinyueZ who is opened another PR)

PR #480, opened 2018-08-20
Update README for latest ktlint version 0.27.0 download-url

XinyueZ:


XinyueZ:
@tiembo Why the test is broken after changing only one piece in README.md?

tiembo:
I'm OK with this change.  Separately, we're exploring why tests are failing on CI.

PR #486, opened 2018-08-28
Making the login test pass

florina-muntenescu:


designernews/src/test/java/io/plaidapp/designernews/ui/login/LoginViewModelTest.kt
```diff
@@ -132,6 +132,10 @@ class LoginViewModelTest {
             enableLoginButton = true
         )
         val uiState = LiveDataTestUtil.getValue(viewModel.uiState)
+        assertEquals(expectedUiModel.showProgress, uiState!!.showProgress)
+        assertEquals(expectedUiModel.showError, uiState.showError)
+        assertEquals(expectedUiModel.showSuccess, uiState.showSuccess)
+        assertEquals(expectedUiModel.enableLoginButton, uiState.enableLoginButton)
         assertEquals(expectedUiModel, uiState)
```
ataulm: if `LoginViewUiModel` is a data class, why isn't this last assert sufficient?

florina-muntenescu: not sure actually. added a todo for more investigation later

PR #487, opened 2018-08-31
Adding the userId to the story response, as long

florina-muntenescu:


core/src/main/java/io/plaidapp/core/designernews/data/stories/model/StoryResponse.kt
```diff
@@ -29,6 +29,7 @@ fun StoryResponse.toStory() = Story(
     commentHtml = comment_html,
     commentCount = comment_count,
     voteCount = vote_count,
+    userId = links?.user ?: 0,
```
ataulm: Is it possible to have a response without a user? 🤔 

In this PR, I would have expected to see more changes (checks for empty string being replaced with checks for `0`). Since there isn't, either:

- a user is always expected, so the ID is always good
- it's ok for the ID to not be unique

if it's the former - maybe using `!!` would be safer, since our assumption will have been proved wrong, but we only made it in one place.

florina-muntenescu: added tests and made the userId and links mandatory in the following commit

PR #492, opened 2018-09-03
Fix warnings in StoryViewHolder

XinyueZ:
Kotlin now sends ```let{}``` warnign when `it` will used directly.


ataulm:
what do the warnings say in this class?

XinyueZ:
The warning is after updating kotlin to 1
60.*(maybe earlier) automatically. There might be some reasons to enable this warning officially. :)

ataulm:
Could you include the warning text? It'd be easier to search and then we could see what the rationale was behind it, if not just explicitness.

ataulm:
https://youtrack.jetbrains.com/issue/KT-11850

Ah it's not clear from the snippet that GitHub shows, but there's a nested lambda in the first instance, so `it` is shadowed.

The remaining two instances shouldn't have flagged as warnings, can you confirm?

XinyueZ:
@ataulm Actually I prefer to keep using explicit names here because all changes there are like this:

```
setOnClickListener { // Thatz a implicit name of `it`
            story?.let { story -> // When another `it` I think the code will be confuse.
                // ....
            }
```
To be honest, if there're more than one nested it-lambda I insist by using explicit namings, well, it's my opinion.

ataulm:
I read it again, my mistake. 💯

XinyueZ:
@ataulm 🤝

XinyueZ:
Promise as well here to rebase later 😏

Issue #493, opened 2018-09-03
Unhealthy in LoginViewModelTest


XinyueZ:
Something not good in this test, please check, i.e not related tickets like #491, #492 cannot go through. 😢  .
```
loginDataChanged_withValidLogin:


java.lang.AssertionError: expected:<true> but was:<false>
	at org.junit.Assert.fail(Assert.java:88)
	at org.junit.Assert.failNotEquals(Assert.java:834)
	at org.junit.Assert.assertEquals(Assert.java:118)
	at org.junit.Assert.assertEquals(Assert.java:144)
	at io.plaidapp.designernews.ui.login.LoginViewModelTest$loginDataChanged_withValidLogin$1.doResume(LoginViewModelTest.kt:139)
	at kotlin.coroutines.experimental.jvm.internal.CoroutineImpl.resume(CoroutineImpl.kt:42)
	at kotlinx.coroutines.experimental.DispatchedTask$DefaultImpls.run(Dispatched.kt:161)
	at kotlinx.coroutines.experimental.DispatchedContinuation.run(Dispatched.kt:25)
	at kotlinx.coroutines.experimental.EventLoopBase.processNextEvent(EventLoop.kt:147)
	at kotlinx.coroutines.experimental.BlockingCoroutine.joinBlocking(Builders.kt:244)
	at kotlinx.coroutines.experimental.BuildersKt.runBlocking(Builders.kt:185)
	at kotlinx.coroutines.experimental.BuildersKt.runBlocking$default(Builders.kt:175)
	at io.plaidapp.designernews.ui.login.LoginViewModelTest.loginDataChanged_withValidLogin(LoginViewModelTest.kt:120)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.junit.runners.model.FrameworkMethod$1.runReflectiveCall(FrameworkMethod.java:50)
	at org.junit.internal.runners.model.ReflectiveCallable.run(ReflectiveCallable.java:12)
	at org.junit.runners.model.FrameworkMethod.invokeExplosively(FrameworkMethod.java:47)
[...]

PR #494, opened 2018-09-04
Remove duplicate/nested launch{} call enableLogin()

XinyueZ:
I'm not sure whether the high probability of test fails #493  is result from this dup call, however, it is real confuse to have nested using of `launch{}` there. The `emitUiState()` has a `launch{}` already internal.


florina-muntenescu:
Good catch!

XinyueZ:
ok, good luck @ CI

PR #498, opened 2018-09-04
Make story.links to StoryLinks? (nullable)

XinyueZ:
because of this
https://github.com/nickbutcher/plaid/blob/781316368c0fefb4774ac26ffd5be6768cbd8eb3/designernews/src/main/java/io/plaidapp/designernews/ui/story/StoryViewModel.kt#L109


florina-muntenescu:
I don't think the `StoryLins` need to be nullable in the `Story`, because Designer News always sends them. Rather, the `?` should be removed from the usage of `story.links`

PR #507, opened 2018-09-07
Remove DesignerNewsPrefs. User related calls replaced with LoginRepos…

florina-muntenescu:
…itory


app/src/main/java/io/plaidapp/ui/HomeActivity.java
```diff
@@ -330,10 +332,11 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 startActivityForResult(ActivityHelper.intentTo(Activities.Search.INSTANCE), RC_SEARCH, options);
                 return true;
             case R.id.menu_designer_news_login:
-                if (!designerNewsPrefs.isLoggedIn()) {
+                if (!loginRepository.isLoggedIn()) {
                     startActivity(ActivityHelper.intentTo(Activities.DesignerNews.Login.INSTANCE));
                 } else {
-                    designerNewsPrefs.logout(HomeActivity.this);
+                    loginRepository.logout();
+                    ShortcutHelper.disablePostShortcut(this);
```
ataulm: there's a corresponding method `enablePostShortcut` but this is unused as far as I can see.

If app shortcuts are enabled by default, then I would expect to see the `enable..` method added in this PR when the user logs in.

If they're not enabled by default, is this call to disable them necessary? (i.e. where is it enabled?)

core/src/main/java/io/plaidapp/core/designernews/data/poststory/PostStoryService.java
```diff
@@ -64,8 +65,11 @@ protected void onHandleIntent(Intent intent) {
         if (intent == null) return;
         if (ACTION_POST_NEW_STORY.equals(intent.getAction())) {
             final boolean broadcastResult = intent.getBooleanExtra(EXTRA_BROADCAST_RESULT, false);
-            final DesignerNewsPrefs designerNewsPrefs = DesignerNewsPrefs.get(this);
-            if (!designerNewsPrefs.isLoggedIn()) return; // shouldn't happen...
+
+            final DesignerNewsService service = Injection.provideDesignerNewsService(this);
+            final LoginRepository repository = Injection.provideLoginRepository(this);
+
+            if (!repository.isLoggedIn()) return; // shouldn't happen...
```
ataulm: `shouldn't happen...` -> `TODO: prompt for login`?

app/src/main/java/io/plaidapp/ui/HomeActivity.java
ataulm: (I see now that it was being called in `DNP.logout()` which is why it's copied here - is it necessary?)

core/src/main/java/io/plaidapp/core/designernews/data/poststory/PostStoryService.java
florina-muntenescu: this is still java. proper functionality will be considered when this is refactored and when designer news actually allows posting a story (the API doesn't seem to be working at the moment)

PR #508, opened 2018-09-09
Making the LoggedInUser a User and using it

florina-muntenescu:
* Creating a LoggedInUserResponse for the login DN API response
* Making the LoggedInUser a User
* Having the LoginRepository, LoginLocalDataSource and LoginRemoteDataSource use LoggedInUser

### DIFF
--- a/core/src/main/java/io/plaidapp/core/designernews/data/login/LoginLocalDataSource.kt
+++ b/core/src/main/java/io/plaidapp/core/designernews/data/login/LoginLocalDataSource.kt
@@ -33,7 +33,7 @@ class LoginLocalDataSource(private val prefs: SharedPreferences) {
             val userId = prefs.getLong(KEY_USER_ID, 0L)
             val username = prefs.getString(KEY_USER_NAME, null)
             val userAvatar = prefs.getString(KEY_USER_AVATAR, null)
-            if (userId == 0L && username == null && userAvatar == null) {
+            if (userId == 0L || username == null) {
                 return null
             }
             // TODO save the entire user
--- a/core/src/main/java/io/plaidapp/core/designernews/data/login/LoginLocalDataSource.kt
+++ b/core/src/main/java/io/plaidapp/core/designernews/data/login/LoginLocalDataSource.kt
@@ -18,7 +18,7 @@ package io.plaidapp.core.designernews.data.login
 
 import android.content.SharedPreferences
 import androidx.core.content.edit
-import io.plaidapp.core.designernews.data.users.model.User
+import io.plaidapp.core.designernews.data.users.model.LoggedInUser
 
 /**
  * Local storage for Designer News login related data, implemented using SharedPreferences
@@ -28,7 +28,7 @@ class LoginLocalDataSource(private val prefs: SharedPreferences) {
     /**
      * Instance of the logged in user. If missing, null is returned
      */
-    var user: User?
+    var user: LoggedInUser?
         get() {
             val userId = prefs.getLong(KEY_USER_ID, 0L)
             val username = prefs.getString(KEY_USER_NAME, null)
@@ -37,12 +37,13 @@ class LoginLocalDataSource(private val prefs: SharedPreferences) {
                 return null
             }
             // TODO save the entire user
-            return User(
+            return LoggedInUser(
                 id = userId,
                 firstName = "",
                 lastName = "",
                 displayName = username,
-                portraitUrl = userAvatar
+                portraitUrl = userAvatar,
+                upvotes = emptyList()
             )
         }
         set(value) {
--- a/core/src/main/java/io/plaidapp/core/designernews/data/users/model/LoggedInUser.kt
+++ b/core/src/main/java/io/plaidapp/core/designernews/data/users/model/LoggedInUser.kt
@@ -18,32 +18,33 @@ package io.plaidapp.core.designernews.data.users.model
 
 import android.arch.persistence.room.ColumnInfo
 import android.arch.persistence.room.Entity
+import android.arch.persistence.room.PrimaryKey
 import com.google.gson.annotations.SerializedName
 
 @Entity(tableName = "logged_in_user")
-class LoggedInUser(
-    id: Long,
-    firstName: String,
-    lastName: String,
-    displayName: String,
-    portraitUrl: String? = null,
+data class LoggedInUser(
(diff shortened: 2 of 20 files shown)


core/src/main/java/io/plaidapp/core/designernews/data/users/model/LoggedInUser.kt
```diff
@@ -18,33 +18,32 @@ package io.plaidapp.core.designernews.data.users.model
...
-data class LoggedInUser(
-    @PrimaryKey
-    @ColumnInfo(name = "id")
-    @SerializedName("id")
-    val id: Long,
+class LoggedInUser(
+    id: Long,
+    firstName: String,
+    lastName: String,
+    displayName: String,
```
ataulm: This could potentially be null - [this code](https://github.com/nickbutcher/plaid/pull/508/files?utf8=%E2%9C%93&diff=split&w=1#diff-b21404c6fdcec0954bd65efb28e62e97R36) allows for it I think; the `&&` could be changed to `||` to make it more explicit what's mandatory and not.

Maybe the linked code needs to change so that it'

core/src/main/java/io/plaidapp/core/designernews/data/login/LoginLocalDataSource.kt
```diff
@@ -37,12 +37,13 @@ class LoginLocalDataSource(private val prefs: SharedPreferences) {
...
             // TODO save the entire user
-            return User(
+            return LoggedInUser(
                 id = userId,
                 firstName = "",
                 lastName = "",
                 displayName = username,
-                portraitUrl = userAvatar
+                portraitUrl = userAvatar,
+                upvotes = emptyList()
```
ataulm: is this the only instance a `LoggedInUser` is created? If so, is it worth ditching the `upvotes` field until this is gonna have values?

core/src/main/java/io/plaidapp/core/designernews/data/users/model/LoggedInUser.kt
```diff
@@ -18,33 +18,32 @@ package io.plaidapp.core.designernews.data.users.model
...
-    @SerializedName("portrait_url")
-    val portraitUrl: String? = null,
+        return true
+    }
 
-    @ColumnInfo(name = "upvotes")
-    @SerializedName("upvotes")
-    val upvotes: List<Long>
-)
+    override fun hashCode(): Int {
```
ataulm: The equals and hashcode looks fishy - how come they're only taking into account the `upvotes`? (which will only ever be empty in this PR, if I read correctly.)

core/src/main/java/io/plaidapp/core/designernews/data/login/LoginLocalDataSource.kt
florina-muntenescu: `LoggedInUser` will actually be saved in the database, but I didn't want to introduce the replacement of `SharedPreferences` with `LoggedInUserDao` in this PR since it can be done separate.

core/src/main/java/io/plaidapp/core/designernews/data/users/model/LoggedInUser.kt
florina-muntenescu: Replaced.

core/src/main/java/io/plaidapp/core/designernews/data/users/model/LoggedInUser.kt
```diff
@@ -18,33 +18,32 @@ package io.plaidapp.core.designernews.data.users.model
...
-import android.arch.persistence.room.PrimaryKey
 import com.google.gson.annotations.SerializedName
 
 @Entity(tableName = "logged_in_user")
-data class LoggedInUser(
-    @PrimaryKey
-    @ColumnInfo(name = "id")
-    @SerializedName("id")
-    val id: Long,
+class LoggedInUser(
```
ataulm: Is it possible to keep this as a data class? (what did this change help us with?)

ataulm: Having LoggedInUser extend from User also couples the two together - since it's a DesignerNews response and not a type we own, probably worth the explicitness over the duplication instead 👍

designernews/src/test/java/io/plaidapp/designernews/TestData.kt
```diff
@@ -32,6 +33,15 @@ import java.util.GregorianCalendar
...
+val loggedInUser = LoggedInUser(
+    id = 111L,
+    firstName = "Plaicent",
+    lastName = "van Plaid",
+    displayName = "Plaicent van Plaid",
+    portraitUrl = "www",
+    upvotes = listOf(123L, 234L, 345L)
+)
+
 val user1 = User(
```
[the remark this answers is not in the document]
florina-muntenescu: it's used in 3 other test classes

core/src/test/java/io/plaidapp/core/designernews/data/login/LoginRepositoryTest.kt
```diff
@@ -36,12 +36,13 @@ class LoginRepositoryTest {
 
     private val username = "user"
     private val pass = "pass"
-    private val user = User(
+    private val user = LoggedInUser(
```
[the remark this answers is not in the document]
florina-muntenescu: Multiple things:
* given that there's only one logged in user defined, and there aren't other dependencies to it, I prefer having the test data inside the test. like this, it's easier to check what was expected and why the test failed, in case of anything. Right now, `loggedInUser` is used in tests where other objects that use properties of the `loggedInUser` are also needed, for example, in `PostStoryCommentUseCaseTest`, we're also building the comment based on the `loggedInUser` data.
* `loggedInUser` is defined in designernews module and we're in core here.

core/src/main/java/io/plaidapp/core/designernews/data/users/model/User.kt
```diff
@@ -21,7 +21,7 @@ import com.google.gson.annotations.SerializedName
 /**
  * Models a Designer News User
  */
-data class User(
+open class User(
```
ataulm: can this be reverted?

florina-muntenescu: wooops. yes.

PR #512, opened 2018-09-11
[WIP] Storing logged in user data in the database instead of shared prefere…

florina-muntenescu:
…nces

### DIFF
--- a/app/src/main/java/io/plaidapp/ui/HomeActivity.java
+++ b/app/src/main/java/io/plaidapp/ui/HomeActivity.java
@@ -98,6 +98,8 @@ import io.plaidapp.ui.recyclerview.FilterTouchHelperCallback;
 import io.plaidapp.ui.recyclerview.GridItemDividerDecoration;
 import io.plaidapp.core.ui.transitions.FabTransform;
 
+import static io.plaidapp.core.AppInjection.provideCoroutinesContextProvider;
+
 public class HomeActivity extends Activity {
 
     private static final int RC_SEARCH = 0;
@@ -335,7 +337,7 @@ public class HomeActivity extends Activity {
                 if (!loginRepository.isLoggedIn()) {
                     startActivity(ActivityHelper.intentTo(Activities.DesignerNews.Login.INSTANCE));
                 } else {
-                    loginRepository.logout();
+                    loginRepository.logout(provideCoroutinesContextProvider());
                     ShortcutHelper.disablePostShortcut(this);
                     // TODO something better than a toast!!
                     Toast.makeText(getApplicationContext(), R.string.designer_news_logged_out,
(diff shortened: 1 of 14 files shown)


app/src/main/java/io/plaidapp/ui/HomeActivity.java
```diff
@@ -335,7 +337,7 @@ public boolean onOptionsItemSelected(MenuItem item) {
                 if (!loginRepository.isLoggedIn()) {
                     startActivity(ActivityHelper.intentTo(Activities.DesignerNews.Login.INSTANCE));
                 } else {
-                    loginRepository.logout();
+                    loginRepository.logout(provideCoroutinesContextProvider());
```
ataulm: Is this something that could be provided as a dependency of the LoginRepository, esp since it's a provider and not the CoroutinesContext itself?

core/src/androidTest/java/io/plaidapp/core/designernews/data/database/LoggedInUserDaoTest.kt
```diff
@@ -34,7 +33,8 @@ class LoggedInUserDaoTest {
     private lateinit var loggedInUser: LoggedInUser
     private lateinit var loggedInUserDao: LoggedInUserDao
 
-    @Before fun setup() {
+    @Before
+    fun setup() {
```
ataulm: is this regulation?

app/src/main/java/io/plaidapp/ui/HomeActivity.java
florina-muntenescu: I didn't fully decide how to handle this, especially given the fact that `getUser` and `logout` are suspending functions. 
Since this is still WIP - you're missing the comments that say that "these methods will be removed once we have a VM for HomeActivity (and others)

core/src/main/java/io/plaidapp/core/designernews/data/login/LoginRepository.kt
```diff
@@ -30,23 +32,41 @@ class LoginRepository(
...
+    suspend fun getUser(): LoggedInUser? {
+        if (user != null) {
+            return user
+        }
+        user = localDataSource.getUser()
+        return user
+    }
+
+    fun getUser(
+        onResult: (user: LoggedInUser?) -> Unit,
```
chrisbanes: Mixing callbacks and coroutines feels wrong (but I get trying to migrate old code)

core/src/main/java/io/plaidapp/core/designernews/data/login/LoginRepository.kt
```diff
@@ -30,23 +32,41 @@ class LoginRepository(
...
+        contextProvider: CoroutinesContextProvider
+    ) = launch(contextProvider.io) {
+        if (user != null) {
```
chrisbanes:
```
(user ?: localDataSource.getUser()).also {
    user = it
    onResult(it)
}
```

chrisbanes: LGTM overall though.

PR #513, opened 2018-09-11
Introduce dagger into :dribbble, :core and parts of :app

keyboardsurfer:
This PR got a little bigger than expected, due to dependencies between :dribbble, :core and :app.
Initially all I wanted to do is add dagger to :dribbble. This is done.
But also parts of :app and :core are now using dagger.

Functions that are injecting without dagger have been marked deprecated.
Also FilterAdapter is now in Kotlin.


core/src/main/java/io/plaidapp/core/ui/FilterAdapter.kt
```diff
@@ -0,0 +1,361 @@
...
+class FilterAdapter @Inject constructor(
+    context: Context,
+    internal val filters: MutableList<Source>
+) : RecyclerView.Adapter<FilterViewHolder>(), FilterSwipeDismissListener {
+
+    private val context: Context = context.applicationContext
+    private var callbacks: MutableList<FiltersChangedCallbacks> = ArrayList()
+
+    val enabledSourcesCount: Int
+        get() = filters.count { it.active }
```
hzsweers: nit: `count(::active)`

core/src/main/java/io/plaidapp/core/ui/FilterAdapter.kt
```diff
@@ -0,0 +1,361 @@
...
+
+    init {
+        setHasStableIds(true)
+    }
+
+    fun getFilters() = filters
```
hzsweers: nit: just make `filters` public if you're not making a defensive copy here?

core/src/main/java/io/plaidapp/core/ui/FilterAdapter.kt
```diff
@@ -0,0 +1,361 @@
...
+    /**
+     * Adds a new data source to the list of filters. If the source already exists then it is simply
+     * activated.
+     *
+     * @param toAdd the source to add
+     * @return whether the filter was added (i.e. if it did not already exist)
+     */
+    fun addFilter(toAdd: Source): Boolean {
+        // first check if it already exists
+        for (i in 0 until filters.size) {
```
hzsweers: could do `filters.find {...}?.let {...}`?

core/src/main/java/io/plaidapp/core/ui/FilterAdapter.kt
```diff
@@ -0,0 +1,361 @@
...
+        notifyItemRemoved(position)
+        dispatchFilterRemoved(removing)
+        SourceManager.removeSource(removing, context)
+    }
+
+    fun getFilterPosition(filter: Source) = filters.indexOf(filter)
+
+    fun enableFilterByKey(key: String, context: Context) {
+        val count = filters.size
+        for (i in 0 until count) {
```
hzsweers: nit: `filters.find {...}`

core/src/main/java/io/plaidapp/core/ui/FilterAdapter.kt
```diff
@@ -0,0 +1,361 @@
...
+    }
+
+    fun unregisterFilterChangedCallback(callback: FiltersChangedCallbacks) {
+        if (!callbacks.isEmpty()) {
+            callbacks.remove(callback)
+        }
+    }
+
+    private fun dispatchFiltersChanged(filter: Source) {
+        if (!callbacks.isEmpty()) {
```
hzsweers: nit: this is a bit unnecessary unless you're trying to optimize for saving the iterator. Same below

app/src/main/java/io/plaidapp/dagger/HomeModule.kt
```diff
@@ -23,20 +23,37 @@ import dagger.Module
...
+ */
+@Module(
+    includes = [
+        DataManagerModule::class,
+        DribbbleDataModule::class,
+        FilterAdapterModule::class,
+        OnDataLoadedModule::class
+    ]
+)
 class HomeModule(private val activity: Activity) {
```
hzsweers: nit: you could switch to `@BindsInstance` for this + make it abstract to make the module completely stateless and avoid instantiation. If you make it `object`, you can make them all JvmStatic too, but see the `@Binds` comment below for tradeoffs consideration

app/src/main/java/io/plaidapp/dagger/HomeModule.kt
```diff
@@ -23,20 +23,37 @@ import dagger.Module
...
 
-    @Provides fun context(): Context = activity
+    @Provides
+    fun context(): Context = activity
```
hzsweers: For this one, if you make this module `abstract` you can just use `@Binds`

```kotlin
@Binds
abstract fun context(activity: Activity): Context
```

app/src/main/java/io/plaidapp/dagger/HomeModule.kt
```diff
@@ -23,20 +23,37 @@ import dagger.Module
...
 
-    @Provides fun activity(): Activity = activity
+    @Provides
+    fun activity(): Activity = activity
```
hzsweers: If you use `@BindsInstance`, this becomes redundant/unnecessary.

app/src/main/java/io/plaidapp/dagger/HomeModule.kt
```diff
@@ -23,20 +23,37 @@ import dagger.Module
...
 
-    @Provides fun columns(): Int = activity.resources.getInteger(R.integer.num_columns)
+    @Provides
+    fun columns(): Int = activity.resources.getInteger(R.integer.num_columns)
 
-    @Provides fun viewPreloadSizeProvider() = ViewPreloadSizeProvider<Shot>()
+    @Provides
+    fun viewPreloadSizeProvider() = ViewPreloadSizeProvider<Shot>()
```
hzsweers: nit: can `ViewPreloadSizeProvider` be made to support constructor injection? Then you don't need to declare here

app/src/main/java/io/plaidapp/dagger/HomeModule.kt
```diff
@@ -23,20 +23,37 @@ import dagger.Module
...
 
-    @Provides fun isPocketInstalled() = PocketUtils.isPocketInstalled(activity)
+    @Provides
+    fun isPocketInstalled() = PocketUtils.isPocketInstalled(activity)
```
hzsweers: nit: this is a potentially expensive call, so maybe require it be injected lazily? Also would definitely recommend qualifying this with something to better distinguish, otherwise on the DI graph this is just `Boolean` and not very obvious what it's in ref to

app/src/main/java/io/plaidapp/dagger/Injector.kt
```diff
@@ -36,11 +37,11 @@ object Injector {
         dataLoadedCallback: BaseDataManager.OnDataLoadedCallback<List<PlaidItem>>
     ) {
         DaggerHomeComponent.builder()
-            .context(activity)
-            .dataLoadedModule(OnDataLoadedModule(dataLoadedCallback))
+            .coreComponent(PlaidApplication.coreComponent(activity))
             .dataManagerModule(DataManagerModule(activity))
```
hzsweers: With all these modules that require an activity, you can remove that dependency since it's provided by the `HomeModule`. Then you can make them all stateless/abstract

app/src/main/java/io/plaidapp/ui/PlaidApplication.kt
```diff
@@ -20,14 +20,20 @@ import android.app.Application
...
 /**
  * Io and Behold
  */
 class PlaidApplication : Application() {
 
     val coreComponent: CoreComponent by lazy {
-        DaggerCoreComponent.create()
+        DaggerCoreComponent
+            .builder()
+            .dribbbleDataModule(DribbbleDataModule())
```
hzsweers: Technically this is unnecessary. Dagger will automatically instantiate modules with default constructors for you

core/src/main/java/io/plaidapp/core/dagger/CoreComponent.kt
```diff
@@ -20,15 +20,22 @@ import dagger.Component
...
-        OnDataLoadedModule::class,
         SharedPreferencesModule::class
     ]
 )
-interface CoreComponent
+interface CoreComponent {
+
+    @Component.Builder interface Builder {
+        fun build(): CoreComponent
+        fun dribbbleDataModule(module: DribbbleDataModule): Builder
```
hzsweers: This shouldn't be necessary since it doesn't require any parameters to instantiate, dagger can handle it for you. Might be off though

core/src/main/java/io/plaidapp/core/dagger/FilterAdapterModule.kt
```diff
@@ -27,5 +28,7 @@ import io.plaidapp.core.ui.FilterAdapter
  */
 @Module class FilterAdapterModule(val context: Context) {
 
-    @Provides fun provideFilterAdapter() = FilterAdapter(context, SourceManager.getSources(context))
+    @Provides fun provideFilterAdapter() = FilterAdapter(context, provideSources())
```
hzsweers: nit: can this be constructor injected?

core/src/main/java/io/plaidapp/core/dagger/FilterAdapterModule.kt
```diff
@@ -27,5 +28,7 @@ import io.plaidapp.core.ui.FilterAdapter
...
+
+    @Provides fun provideSources(): MutableList<Source> = SourceManager.getSources(context)
```
hzsweers: nit: exposing a mutable list on the DI graph is asking for a bad time. Also, Dagger won't distinguish anyway since it doesn't see a difference

core/src/main/java/io/plaidapp/core/dagger/FilterAdapterModule.kt
```diff
@@ -27,5 +28,7 @@ import io.plaidapp.core.ui.FilterAdapter
  */
 @Module class FilterAdapterModule(val context: Context) {
```
hzsweers: nit: can't this context be provided by the HomeModule?

core/src/main/java/io/plaidapp/core/dagger/OnDataLoadedModule.kt
```diff
@@ -24,7 +24,10 @@ import io.plaidapp.core.data.PlaidItem
 /**
  * Module to provide [BaseDataManager.OnDataLoadedCallback].
  */
-@Module class OnDataLoadedModule(private val callback: BaseDataManager.OnDataLoadedCallback<List<PlaidItem>>) {
+@Module class OnDataLoadedModule(
```
hzsweers: nit: if you're not reusing this module, you could just move this to be a `@BindsInstance`-provided value in your home component

core/src/main/java/io/plaidapp/core/dagger/dribbble/DribbbleDataModule.kt
```diff
@@ -16,14 +16,53 @@
...
+    @Provides fun provideShotsRepository(
+        remoteDataSource: SearchRemoteDataSource,
+        contextProvider: CoroutinesContextProvider
+    ) = ShotsRepository.getInstance(remoteDataSource, contextProvider)
+
+    @Provides fun provideDribbbleSearchService(): DribbbleSearchService =
+        provideRetrofit().create(DribbbleSearchService::class.java)
+
+    @Provides fun provideRetrofit(): Retrofit {
+        return Retrofit.Builder()
```
hzsweers: Also, here, if you provide it lazily, you can defer it initialization to off the UI thread on whatever coroutine worker does the call (see my other comment).

core/src/main/java/io/plaidapp/core/dagger/dribbble/DribbbleDataModule.kt
```diff
@@ -16,14 +16,53 @@
...
+            .baseUrl(DribbbleSearchService.ENDPOINT)
+            .addConverterFactory(DribbbleSearchConverter.Factory())
+            .addCallAdapterFactory(CoroutineCallAdapterFactory())
+            .client(provideOkHttpClient(provideLoggingInterceptor()))
+            .build()
+    }
+
+    @Provides fun provideLoggingInterceptor(): HttpLoggingInterceptor {
```
hzsweers: nit: A third logging interceptor provider?

core/src/main/java/io/plaidapp/core/dagger/dribbble/DribbbleDataModule.kt
```diff
@@ -16,14 +16,53 @@
...
+        return HttpLoggingInterceptor().apply {
+            level = if (BuildConfig.DEBUG) {
+                HttpLoggingInterceptor.Level.BODY
+            } else {
+                HttpLoggingInterceptor.Level.NONE
+            }
+        }
+    }
+
+    @Provides fun provideOkHttpClient(loggingInterceptor: HttpLoggingInterceptor): OkHttpClient =
```
hzsweers: This will create a new `OkHttpClient` from the one created in your other data module, which would have its own unique connection pool (possible memory consumption issue). You likely at least want to reuse the one provided in your core data module, then call `.newBuilder()` on it to create a new instance while sharing the connection pool. Then you can also share the logging interceptor under the hood

core/src/main/java/io/plaidapp/core/dagger/OkHttpModule.kt
```diff
@@ -0,0 +1,42 @@
...
+import dagger.Module
+import dagger.Provides
+import io.plaidapp.core.BuildConfig
+import okhttp3.OkHttpClient
+import okhttp3.logging.HttpLoggingInterceptor
+
+/**
+ * Module to provide [OkHttpClient] and [HttpLoggingInterceptor].
+ */
+@Module class OkHttpModule {
```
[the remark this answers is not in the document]
keyboardsurfer: Same as LoggingInterceptorModule

core/src/main/java/io/plaidapp/core/ui/FilterAdapter.kt
keyboardsurfer: Done

core/src/main/java/io/plaidapp/core/dagger/dribbble/DribbbleDataModule.kt
keyboardsurfer: That's the one that remained.

core/src/main/java/io/plaidapp/core/ui/FilterAdapter.kt
keyboardsurfer: Thought the same, but the position is being used within `notifyItemChanged` below.

hzsweers: :+1:

PR #519, opened 2018-09-28
Converting search to a dynamic feature and making it depend on app, i…

florina-muntenescu:
…nstead of having app depend on it


search/build.gradle
```diff
@@ -37,5 +37,5 @@ repositories {
 }
 
 dependencies {
-    implementation project(':core')
+    implementation project(':app')
```
rcd27: Doesn't that need `api project(':app')` in order to compile it each time in case sources in `app` have changed?

keyboardsurfer: `api` is used when the module depending on `:app` would want to expose it further.
This is not the case, so `implementation` is correct.

PR #528, opened 2018-10-20
Using Kotlin's structured concurrency model

florina-muntenescu:
ViewModels have a scope and the parent job gets cancelled in `onCleared`
Repositories and UseCases, that are still called from Java and use the mix of coroutines and callbacks, since they also offer a way of cancelling the tasks, are using the `GlobalScope` to launch coroutines.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -24,7 +24,7 @@ buildscript {
             'constraintLayout'  : '2.0.0-alpha2',
             'coreKtx'           : '1.0.0',
             'androidxCollection': '1.0.0',
-            'coroutines'        : '0.30.0-eap13',
+            'coroutines'        : '1.0.0-RC1',
             'crashlytics'       : '2.9.4',
             'dagger'            : '2.16',
             'espresso'          : '3.1.0-beta02',
--- a/build.gradle
+++ b/build.gradle
@@ -24,7 +24,7 @@ buildscript {
             'constraintLayout'  : '2.0.0-alpha2',
             'coreKtx'           : '1.0.0',
             'androidxCollection': '1.0.0',
-            'coroutines'        : '0.26.1-eap13',
+            'coroutines'        : '0.30.0-eap13',
             'crashlytics'       : '2.9.4',
             'dagger'            : '2.16',
             'espresso'          : '3.1.0-beta02',
--- a/core/src/main/java/io/plaidapp/core/designernews/domain/LoadStoriesUseCase.kt
+++ b/core/src/main/java/io/plaidapp/core/designernews/domain/LoadStoriesUseCase.kt
@@ -22,6 +22,7 @@ import io.plaidapp.core.data.Result
 import io.plaidapp.core.data.prefs.SourceManager
 import io.plaidapp.core.designernews.data.stories.StoriesRepository
 import io.plaidapp.core.designernews.data.stories.model.toStory
+import kotlinx.coroutines.GlobalScope
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.withContext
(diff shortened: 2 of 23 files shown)


core/src/main/java/io/plaidapp/core/designernews/domain/LoadStoriesUseCase.kt
```diff
@@ -44,16 +45,16 @@ class LoadStoriesUseCase(
         page: Int,
         callback: LoadSourceCallback,
         jobId: String
-    ) = launch(contextProvider.io) {
+    ) = GlobalScope.launch(dispatcherProvider.computation) {
```
chrisbanes: You probably don't want GlobalScope. Wrap it in a `coroutineScope {}` call.

core/src/main/java/io/plaidapp/core/designernews/domain/SearchStoriesUseCase.kt
```diff
@@ -44,16 +45,16 @@ class SearchStoriesUseCase(
         page: Int,
         callback: LoadSourceCallback,
         jobId: String
-    ) = launch(contextProvider.io) {
+    ) = GlobalScope.launch(dispatcherProvider.computation) {
```
chrisbanes: Ditto about GlobalScope

designernews/src/main/java/io/plaidapp/designernews/ui/login/LoginViewModel.kt
```diff
@@ -54,36 +58,39 @@ class LoginViewModel(
...
-        val result = loginRepository.login(username, password)
-
-        if (result is Result.Success) {
-            val user = result.data
-            emitUiState(
-                showSuccess = Event(
-                    LoginResultUiModel(
-                        user.displayName.toLowerCase(),
-                        user.portraitUrl
+    private fun launchLogin(username: String, password: String) =
```
chrisbanes: Nit: use block style if the line goes over 1 line

florina-muntenescu: done

core/src/main/java/io/plaidapp/core/designernews/domain/SearchStoriesUseCase.kt
florina-muntenescu: done

florina-muntenescu:
@chrisbanes reading for another 👀

build.gradle
```diff
@@ -23,7 +23,7 @@ buildscript {
             'androidx'          : '1.0.0',
             'constraintLayout'  : '2.0.0-alpha2',
             'coreKtx'           : '1.0.0',
-            'coroutines'        : '0.26.1-eap13',
+            'coroutines'        : '0.30.0-eap13',
```
chrisbanes: There's 0.30.2-eap13 now (or even 1.0.0-RC1)

core/src/main/java/io/plaidapp/core/dribbble/data/ShotsRepository.kt
```diff
@@ -63,13 +64,13 @@ class ShotsRepository constructor(
         page: Int,
         id: String,
         onResult: (Result<List<Shot>>) -> Unit
-    ) = launch(contextProvider.io) {
+    ) = GlobalScope.launch(dispatcherProvider.io) {
```
chrisbanes: Sure you want to use GlobalScope? The standard way would be to do:

``` kotlin
private fun launchSearch(...) = coroutineScope {
    launch(dispatcherProvider.io) {
        // ...
    }
}
```

florina-muntenescu: `coroutineScope` doesn't work with the mix of coroutines and callbacks we have here (due to `search` being called from Java). 
But, I replaced the GlobalScope with one defined in this class.

PR #453, opened 2018-07-31
Move more logic into binding adapters.

nickbutcher:

PR #529, opened 2018-10-21
Cleaning up the nullability types in dribbble data classes.

florina-muntenescu:
Splitting the code from #453 into multiple PRs, to avoid rebasing and, to be easily reviewed.


core/src/main/java/io/plaidapp/core/dribbble/data/api/model/User.kt
```diff
@@ -25,10 +25,10 @@ data class User(
     @SerializedName("id") val id: Long,
     @SerializedName("name") val name: String,
     @SerializedName("username") val username: String,
-    @SerializedName("avatar_url") val avatarUrl: String? = null
+    @SerializedName("avatar_url") val avatarUrl: String
```
ataulm: dribbble v1 said `Blank fields are included as null instead of being omitted.`
this warning isn't available on the v2 api as far as I could see - is this safe?

If the field is omitted, what's the behaviour of the json parser?

core/src/main/java/io/plaidapp/core/dribbble/data/api/model/Images.kt
```diff
@@ -20,16 +20,17 @@ import io.plaidapp.core.dribbble.data.api.model.Images.ImageSize.NORMAL_IMAGE_SI
 import io.plaidapp.core.dribbble.data.api.model.Images.ImageSize.TWO_X_IMAGE_SIZE
 
 /**
- * Models links to the various quality of images of a shot.
+ * Models links to the various quality of images of a shot. One of [hidpi] or [normal] must be
+ * non-null.
```
ataulm: and non-empty? The check below for `!hidpi.isNullOrEmpty()` suggests empty is an invalid state - should this be checked on the normal one too, throwing an exception? If it's not a big deal that it's empty (i.e. we assume it will either be null or valid), should the check below be relaxed to `hidpi != null`?

dribbble/src/main/java/io/plaidapp/dribbble/domain/ImageUriProvider.kt
```diff
@@ -40,12 +39,12 @@ class ImageUriProvider @Inject constructor(context: Context) {
      * extension-less name, and then get the Uri.
      */
     @Suppress("RedundantSuspendModifier")
-    suspend operator fun invoke(url: String, size: Images.ImageSize): Uri {
+    suspend operator fun invoke(url: String, width: Int, height: Int): Uri {
```
ataulm: 👍 should this class belong in a different package/module? It looks like this change was done to stop relying on dribbble data classes?

core/src/main/java/io/plaidapp/core/dribbble/data/api/model/User.kt
florina-muntenescu: Looks like the field is never omitted. I just looked for a user that doesn't have an avatar set and dribbble returns the link to the default image.

core/src/main/java/io/plaidapp/core/dribbble/data/api/model/Images.kt
florina-muntenescu: done

ataulm:
👌

dribbble/src/main/java/io/plaidapp/dribbble/domain/ImageUriProvider.kt
florina-muntenescu: yeah. Refactored it to pass the file authority and now it's in core.utils.

florina-muntenescu:
@ataulm - thanks for the initial review. Given that the fix for one of your comments wasn't trivial, another 👀 would be appreciated

core/src/main/java/io/plaidapp/core/dribbble/data/api/model/Images.kt
```diff
@@ -30,11 +30,11 @@ data class Images(
 ) {
 
     fun best(): String {
-        return if (!hidpi.isNullOrEmpty()) hidpi else normal!!
+        return hidpi ?: normal!!
```
ataulm: 👍

core/src/main/java/io/plaidapp/core/dribbble/data/api/model/Images.kt
```diff
@@ -30,11 +30,11 @@ data class Images(
...
     }
 
     fun bestSize(): ImageSize {
-        return if (!hidpi.isNullOrEmpty()) TWO_X_IMAGE_SIZE else NORMAL_IMAGE_SIZE
+        return if (hidpi != null) TWO_X_IMAGE_SIZE else NORMAL_IMAGE_SIZE
```
ataulm: 👍

dribbble/src/main/java/io/plaidapp/dribbble/dagger/DribbbleModule.kt
```diff
@@ -65,4 +66,7 @@ class DribbbleModule(private val activity: ShotActivity, private val shotId: Lon
             coroutinesContextProvider
         )
     }
+
+    @Provides
+    fun fileAuthority() = BuildConfig.FILES_AUTHORITY
```
ataulm: As far as Dagger is concerned, `ImageUriProvider` is looking for a `String`, right? Is this fragile?

core/src/main/java/io/plaidapp/core/dribbble/data/api/model/Images.kt
rcd27: You have Kotlin 1.3 now and this is good place to use (Contracts)[https://kotlinlang.org/docs/reference/whatsnew13.html#contracts] :)

dribbble/src/main/java/io/plaidapp/dribbble/dagger/DribbbleModule.kt
florina-muntenescu: @keyboardsurfer what do you think?

core/src/main/java/io/plaidapp/core/dribbble/data/api/model/Images.kt
florina-muntenescu: I know, right! I've been thinking about applying Contracts in Plaid for a while but I think we'll do it in a future, separate PR. Plus, I should leave @nickbutcher some fun Kotlin stuff to play with also.

dribbble/src/main/java/io/plaidapp/dribbble/dagger/DribbbleModule.kt
florina-muntenescu: done!

florina-muntenescu: done!

ataulm:
:ok_hand::tada:

PR #539, opened 2018-10-27
Demo BindsInstance + static provides 

### LINKED ISSUE #538: [Dagger] Use the object + @JvmStatic + @BindsInstance pattern
Follow up to Ben's [PR](https://github.com/nickbutcher/plaid/pull/520)

hzsweers:
Resolves #538

Opportunistically converted `SearchDataManager` to clean up generics across java interop and eliminate the need for the manual `@Provides` for it in the module

Now `SearchModule` is fully stateless and also not instantiated at runtime or part of the component builder. If the other modules it includes were moved to this pattern as well, they could all piggyback off of the exposed `Context` and `Activity` depenendecies and create a fully stateless graph


ataulm:
>Opportunistically converted SearchDataManager to clean up generics across java interop and eliminate the need for the manual @Provides for it in the module

can you explain this bit further please? I was reading the diff, I couldn't comprehend why/how there's no explicit `provides`.

hzsweers:
No prob :). 

Also - Bit of trivia, but I suppose `SearchModule` technically _is_ instantiated as a singleton at runtime. If you really really really wanted to avoid the object allocation, you'd have to do an awkward `abstract` class + companion object and then let proguard/R8 strip out the companion object for you

```kotlin
@Module
abstract class TacoModule {
  @Module
  companion object {
    @Provides
    @JvmStatic
    fun provideTaco() = Taco()
  }
}
```

To dagger, this looks like this in java

```java
@Module
abstract class TacoModule {
  
  @Provides
  static Taco provideTaco() {
    return new Taco();
  }

  @Module
  static class Companion {
     @Provides
     static Taco provideTaco() {
       return new Taco();
     }
  }
}
```

The `provideTaco` function actually gets duplicated by the Kotlin compiler into the enclosing `TacoModule` class. Dagger would use this synthetic copied one (so the one in the companion object would look unused in the IDE), but also generate unused factory code for the `Companion` module (something like `TacoModule$Companion_TacoFactory`) that R8 would hopefully remove since that "companion module" isn't included/used anywhere. Hope that makes sense

That said, the allocation is mostly harmless unless you're flattening dozens of modules and worried about classloading at runtime, and dagger is smart enough to notice when your module has all static provides methods (even if it's not abstract). I think I prefer having the code be easier to read :)

hzsweers:
The CI failure looks unrelated o_O

ataulm:
```kotlin
class Taco @Inject constructor(private val seasoning: Seasoning) {
}
```

I have something similar, where `Seasoning` is an interface. `Salt` is an implementation of `Seasoning` which Dagger is aware how to create:

```kotlin
interface Seasoning

class Salt @Inject constructor(private val somethingProvided: SomethingProvided) : Seasoning
```

With this, I get an error: 

```kotlin
class Taco @Inject constructor(private val seasoning: Seasoning)
```

```
e: ...paintings/PaintingsComponent.java:9: error: [Dagger/MissingBinding]
com.ataulm.artcollector.paintings.domain.PaintingsRepository
cannot be provided without an @Provides-annotated method.
    public abstract void inject(@org.jetbrains.annotations.NotNull()
...
```

(where PaintingsRepository is equivalent to the `Seasoning` interface).

but with this, I don't:

```kotlin
class Taco @Inject constructor(private val salt: Salt)
```

tiembo:
Re-ran CI - it's green now! :heavy_check_mark:

florina-muntenescu:
@hzsweers thanks for the PR. If it's ok with you, we'll try to merge #520 first since Ben has been working on this for a while, and then I'll ask you to rebase.

hzsweers:
👍 sounds good

florina-muntenescu:
@hzsweers could you rebase, please?

hzsweers:
Yep will take a look this weekend!

hzsweers:
Updated!

PR #526, opened 2018-10-16
[ISSUE] Upgrading to gradle 3.3.0-alpha10 causes a dependency resolve error

florina-muntenescu:
PR created for this [gradle issue](https://issuetracker.google.com/117808473)

PR #540, opened 2018-10-27
AGP 3.3.0-beta-01 with suggested classpath fixes

hzsweers:
Resolves the issue raised in #526 with the recommended solution from the issue


tiembo:
Thanks @hzsweers !

PR #541, opened 2018-10-28
Moving creating the comment reply focus gained and lost animator to t…

florina-muntenescu:
…he CommentReplyViewHolder

### DIFF
--- a/designernews/src/main/java/io/plaidapp/designernews/ui/story/CommentReplyViewHolder.kt
+++ b/designernews/src/main/java/io/plaidapp/designernews/ui/story/CommentReplyViewHolder.kt
@@ -16,15 +16,21 @@
 
 package io.plaidapp.designernews.ui.story
 
-import com.google.android.material.textfield.TextInputLayout
-import androidx.recyclerview.widget.RecyclerView
+import android.animation.Animator
+import android.animation.AnimatorSet
+import android.animation.ObjectAnimator
+import android.animation.PropertyValuesHolder
 import android.view.View
 import android.widget.Button
 import android.widget.EditText
 import android.widget.ImageButton
-
-import io.plaidapp.designernews.R
+import androidx.core.animation.doOnEnd
+import androidx.core.animation.doOnStart
+import androidx.recyclerview.widget.RecyclerView
+import com.google.android.material.textfield.TextInputLayout
 import io.plaidapp.core.designernews.domain.model.Comment
+import io.plaidapp.core.util.AnimUtils.getFastOutSlowInInterpolator
+import io.plaidapp.designernews.R
 
 /**
  * View holder for a Designer News comment reply.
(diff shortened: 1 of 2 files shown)


designernews/src/main/java/io/plaidapp/designernews/ui/story/CommentReplyViewHolder.kt
```diff
@@ -33,12 +37,73 @@ import io.plaidapp.core.designernews.domain.model.Comment
...
+            .interpolator = interpolator
+
+        postReply.visibility = View.VISIBLE
+        postReply.alpha = 0f
+        postReply.animate()
+            .alpha(1f)
+            .setDuration(200L)
+            .interpolator = interpolator
+
+        return AnimatorSet().apply {
```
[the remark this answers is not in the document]
nickbutcher: +1 if we want to return an `Animator` (I assume so that we can move to a proper `ItemAnimator` at some point, controlled externally) then I then I don't think we can use an empty set but should probably switch from `ViewPropertyAnimator` to a set of `Animator`s. e.g.

```
val commentVotesAnim = ObjectAnimator.ofPropertyValuesHolder(commentVotes, 
    PVH.ofFloat(View.TRANSLATION_X, -commentVotes.width),
    PVH.of ...

val replyLabelAnim = ...
val postReplyAnim = ...

return AnimatorSet().playTogether(commentVotesAnim, replyLabelAnim, postReplyAnim).also {
    doOnStart ...
}
```

designernews/src/main/java/io/plaidapp/designernews/ui/story/CommentReplyViewHolder.kt
```diff
@@ -33,12 +39,88 @@ import io.plaidapp.core.designernews.domain.model.Comment
...
+                postReply,
+                PropertyValuesHolder.ofFloat(View.ALPHA, 0f))
+        postReplyAnimator.duration = 200L
+        postReplyAnimator.interpolator = interpolator
+
+        val animator = AnimatorSet()
+        animator.playTogether(commentVotesAnimator, replyLabelAnimator, postReplyAnimator)
+        animator.apply {
+            doOnStart {
+                itemView.setHasTransientState(true)
```
ataulm: @nickbutcher the transient state flag/counter is set but never checked in the app. I read that it's used by RecyclerView to check whether a view (that's no longer a child of the RV) is allowed to be recycled; if it has transient state, it's not allowed and `RecyclerView.Adapter#onFailedToRecycleView(): Boolean` will be called, but this function is not overridden in Plaid.

Is there any intrinsic benefit to setting the transient state without overriding that function in the Adapter (where it might for example cancel the animation)?

I read also that if the animation was created by the `RecyclerView.ItemAnimator`, then that callback will never be triggered because the RecyclerView will keep the view as a child until the animation has completed. Is this change going to facilitate the removal of having to handle `setHasTransientState()` manually or it'll still be necessary?

nickbutcher:
Sounds like a plan.

designernews/src/main/java/io/plaidapp/designernews/ui/story/CommentReplyViewHolder.kt
```diff
@@ -33,12 +39,92 @@ import io.plaidapp.core.designernews.domain.model.Comment
...
+
+        val commentVotesAnimator = ObjectAnimator.ofPropertyValuesHolder(
+                commentVotes,
+                PropertyValuesHolder.ofFloat(View.TRANSLATION_X, -commentVotes.width.toFloat()),
+                PropertyValuesHolder.ofFloat(View.ALPHA, 0f)
+        ).apply {
+            duration = 200L
+            this.interpolator = interpolator
+        }
+        val replyLabelAnimator = ObjectAnimator.ofPropertyValuesHolder(
```
nickbutcher: We only need to use `PropertyValuesHolder` if we're animating multiple properties of the same view. So here (and others) we should just `ObjectAnimator.ofFloat`.

PR #542, opened 2018-10-29
Upgrade Kotlin and coroutines to stable

florina-muntenescu:


hzsweers:
> Can probably clean up the kotlinx maven repos while you're at it :)

Ignore me I'm imagining things

PR #545, opened 2018-11-02
Move the loading displaying in Login, on the UI thread

florina-muntenescu:


designernews/src/main/java/io/plaidapp/designernews/ui/login/LoginViewModel.kt
```diff
@@ -70,7 +70,7 @@ class LoginViewModel @Inject constructor(
             if (!isLoginValid(username, password)) {
                 return@launch
             }
-            showLoading()
+            withContext(dispatcherProvider.main) { showLoading() }
```
ataulm: how didn't this fail before? It's using `setValue` 🤔

florina-muntenescu: the computation dispatcher was only added once I added the structured concurrency.

PR #551, opened 2018-11-14
Fix crash in UpvoteService due to incorrect injection

### LINKED ISSUE #550: [Bug] Crash on UpvoteStoryService.handleActionUpvote
```
Fatal Exception: java.lang.NullPointerException
Attempt to invoke virtual method 'boolean io.plaidapp.core.designernews.data.login.LoginRepository.isLoggedIn()' on a null object reference
```

Looks like we're not injecting the LoginRepository correctly.

florina-muntenescu:
Fixes #550

### DIFF
--- a/core/src/main/java/io/plaidapp/core/dagger/designernews/Injector.kt
+++ b/core/src/main/java/io/plaidapp/core/dagger/designernews/Injector.kt
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2018 Google, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


core/src/main/java/io/plaidapp/core/designernews/data/votes/UpvoteStoryService.java
```diff
@@ -36,6 +37,7 @@
 
     public UpvoteStoryService() {
         super("UpvoteStoryService");
+        Injector.inject(this);
```
ataulm: yep, that seems like the issue.

core/src/main/java/io/plaidapp/core/dagger/designernews/Injector.kt
```diff
@@ -0,0 +1,45 @@
...
+import io.plaidapp.core.designernews.data.api.DesignerNewsService
+import io.plaidapp.core.designernews.data.login.LoginLocalDataSource
+import io.plaidapp.core.designernews.data.votes.UpvoteStoryService
+import retrofit2.converter.gson.GsonConverterFactory
+
+/**
+ * Injector for [UpvoteStoryService].
+ */
+
+fun inject(service: UpvoteStoryService) {
```
ataulm: Would it be easy to move UpvoteStoryService to `designernews` instead of fixing this stuff in `core`?

PR #556, opened 2018-11-27
Use implementation instead of api

keyboardsurfer:
Gradle figures out where stuff is needed and where to place code in the app bundle.
This also can make incremental builds faster.


ataulm:
:+1: I think there could be some value in moving these two `apply from ..` lines next the `dependencies` blocks.

PR #557, opened 2018-11-28
Fixing dribbble parsing issue that allows the search to run.

florina-muntenescu:
Dribbble changed their HTML for likes. Also, now, the likes tag can be missing completely in case the shot doesn't have any likes.


florina-muntenescu:
Fixes DribbbleSearchConverter crash from Crashlytics

florina-muntenescu:
GoogleBot doesn't seem to like that a commit was co-authored so manually turned the CLA flag to yes.

PR #558, opened 2018-11-28
Cleaning a bit the search module and converting files to Kotlin

florina-muntenescu:
This branch was rebased on top of [fm/search_not_working](https://github.com/nickbutcher/plaid/tree/fm/search_not_working) fix, so it can be actually tested.


tiembo:
CI is now green

PR #565, opened 2018-11-28
Update Gradle wrapper JAR and gradlew

tiembo:
via running `./gradlew wrapper --gradle-version 5.0 --distribution-type all`


tiembo:
The memory settings were automatically applied when running the `./gradlew wrapper` command, which reflects the new defaults: https://docs.gradle.org/5.0/userguide/upgrading_version_4.html#rel5.0:default_memory_settings

PR #566, opened 2018-11-30
[WIP] Update memory setting for CircleCI

tiembo:


tiembo:
Yes - improvements suggested by @chrisbanes.  Paraphrasing:

* JAVA_TOOL_OPTIONS is picked up by all JVMs
* 1024M is important when moving to R8.  2x workers + R8 at 1.5G would exceed max ram

PR #578, opened 2018-12-27
Removes BaseDataManager

florina-muntenescu:
Removes `BaseDataManager` so we can have the main feed and the search independent from each other, so we can have more flexibility in the way we're implementing both features.


search/src/main/java/io/plaidapp/search/domain/SearchDataManager.kt
```diff
@@ -34,20 +36,22 @@ import javax.inject.Inject
  */
 class SearchDataManager @Inject constructor(
     context: Context,
-    onDataLoadedCallback: BaseDataManager.OnDataLoadedCallback<List<PlaidItem>>,
+    private var onDataLoadedCallback: OnDataLoadedCallback<List<PlaidItem>>,
```
ataulm: This can be `private val`

core/src/main/java/io/plaidapp/core/data/DataManager.java
```diff
@@ -67,13 +73,20 @@ public DataManager(OnDataLoadedCallback<List<? extends PlaidItem>> onDataLoadedC
         inflightCalls = new HashMap<>();
     }
 
+    public void setOnDataLoadedCallback(OnDataLoadedCallback<List<? extends PlaidItem>> onDataLoadedCallback) {
+        this.onDataLoadedCallback = onDataLoadedCallback;
+    }
+
+    private final void onDataLoaded(List<? extends PlaidItem> data) {
```
ataulm: if this is private, can the body be inlined?

ataulm: I can't tell if I'm blind, but is this method called in this class?

search/src/main/java/io/plaidapp/search/domain/SearchDataManager.kt
```diff
@@ -34,20 +36,22 @@ import javax.inject.Inject
...
 
     // state
     var query = ""
         private set
     private var page = 1
 
-    init {
-        setOnDataLoadedCallback(onDataLoadedCallback)
-        searchStories = provideSearchStoriesUseCase(context)
+    fun onDataLoaded(data: List<PlaidItem>) {
```
ataulm: Is anyone outside this class calling this? If not, can this be inlined?

search/src/main/java/io/plaidapp/search/domain/SearchDataManager.kt
```diff
@@ -34,20 +36,22 @@ import javax.inject.Inject
...
     context: Context,
-    onDataLoadedCallback: BaseDataManager.OnDataLoadedCallback<List<PlaidItem>>,
+    private var onDataLoadedCallback: OnDataLoadedCallback<List<PlaidItem>>,
     private val shotsRepository: ShotsRepository
-) : BaseDataManager<List<PlaidItem>>(), LoadSourceCallback {
+) : DataLoadingSubject, LoadSourceCallback {
 
-    private val searchStories: SearchStoriesUseCase
+    private val loadingCount: AtomicInteger = AtomicInteger(0)
+    private var loadingCallbacks: MutableList<DataLoadingSubject.DataLoadingCallbacks>? = null
```
ataulm: Is there much value in making this nullable? I expect there will be at least one callback at one time (super assumption 😁 ) and even if there isn't, an empty list allocation is worth removing the noise of a nullable var? 

`private val loadingCallbacks: List<DataLoadingSubject.DataLoadingCallbacks> = mutableListOf()`

search/src/main/java/io/plaidapp/search/dagger/Injector.kt
```diff
@@ -36,7 +36,7 @@ object Injector {
     @JvmStatic
     fun inject(
         activity: SearchActivity,
-        dataLoadedCallback: BaseDataManager.OnDataLoadedCallback<List<PlaidItem>>
+        dataLoadedCallback: OnDataLoadedCallback<List<PlaidItem>>
```
hzsweers: Random thought for the future once it's 100% kotlin - if you did 

```kotlin
onDataLoadedCallback: (data: List<PlaidItem>) -> Unit
```

That would let you write more idiomatic kotlin code in places that use this, like in SearchActivity

```kotlin
Injector.inject(this) { data ->

}
```

I'm still figuring out when to use named interfaces vs just functions, but currently my rule of thumb is basically 

```
- do other types implement it or is it on the DI graph?
  - Use an interface
- is it only ever used anonymously?
  - Use higher order function with named param
```

core/src/main/java/io/plaidapp/core/data/DataManager.java
```diff
@@ -30,34 +30,40 @@
...
 
     private final ShotsRepository shotsRepository;
     private final FilterAdapter filterAdapter;
     private final LoadStoriesUseCase loadStoriesUseCase;
     private final SearchStoriesUseCase searchStoriesUseCase;
     private final ProductHuntRepository productHuntRepository;
-    Map<String, Integer> pageIndexes;
-    Map<String, Call> inflightCalls;
+    private Map<String, Integer> pageIndexes;
+    private Map<String, Call> inflightCalls;
```
hzsweers: Random nit - I noticed that this never has any calls put into it, only queried/removed.

core/src/main/java/io/plaidapp/core/data/DataManager.java
```diff
@@ -30,34 +30,40 @@
...
 
     public DataManager(OnDataLoadedCallback<List<? extends PlaidItem>> onDataLoadedCallback,
                        LoadStoriesUseCase loadStoriesUseCase,
                        SearchStoriesUseCase searchStoriesUseCase,
                        ShotsRepository shotsRepository,
                        FilterAdapter filterAdapter) {
         super();
+        this.loadingCount = new AtomicInteger(0);
```
hzsweers: nit: can inline this to the property itself. Same with a few others. Makes it clearer what this constructor is actually initializing

search/src/main/java/io/plaidapp/search/domain/SearchDataManager.kt
hzsweers: +1, especially if it's an injected property

search/src/main/java/io/plaidapp/search/domain/SearchDataManager.kt
```diff
@@ -100,12 +105,65 @@ class SearchDataManager @Inject constructor(
...
+                Source.DesignerNewsSearchSource.DESIGNER_NEWS_QUERY_PREFIX + query
+            )
             onDataLoaded(result)
         }
     }
 
     override fun loadFailed(source: String) = loadFinished()
+
+    override fun isDataLoading(): Boolean {
+        return loadingCount.get() > 0
```
hzsweers: nit: body can be inlined

`override fun isDataLoading() = loadingCount.get() > 0`

Same with a few other single-line bodies below

search/src/main/java/io/plaidapp/search/domain/SearchDataManager.kt
```diff
@@ -35,20 +36,22 @@ import javax.inject.Inject
...
 class SearchDataManager @Inject constructor(
     context: Context,
-    onDataLoadedCallback: OnDataLoadedCallback<List<PlaidItem>>,
+    private var onDataLoadedCallback: OnDataLoadedCallback<List<PlaidItem>>,
     private val shotsRepository: ShotsRepository
-) : BaseDataManager<List<PlaidItem>>(), LoadSourceCallback {
+) : DataLoadingSubject, LoadSourceCallback {
 
-    private val searchStories: SearchStoriesUseCase
+    private val loadingCount: AtomicInteger = AtomicInteger(0)
```
hzsweers: Do you expect this to be hit from multiple threads? Same with the one in `DataManager`

search/src/main/java/io/plaidapp/search/dagger/Injector.kt
florina-muntenescu: In the future, there will be no callback like this. The role of this PR is just to separate the DataManager and the SearchDataManager so I can then completely remove them and replace them with something else.

search/src/main/java/io/plaidapp/search/domain/SearchDataManager.kt
florina-muntenescu: It's just copied for now. Won't change it.

search/src/main/java/io/plaidapp/search/domain/SearchDataManager.kt
florina-muntenescu: making this nonnull but leaving it as a list. to be changed to something proper in another PR.

core/src/main/java/io/plaidapp/core/data/DataManager.java
florina-muntenescu: ha! true. I'll leave this as it is for now, proper cleanup/rewrite to happen later.

florina-muntenescu:
@hzsweers and @ataulm ready for another 👀

search/src/main/java/io/plaidapp/search/domain/SearchDataManager.kt
```diff
@@ -34,21 +36,21 @@ import javax.inject.Inject
...
     // state
     var query = ""
         private set
     private var page = 1
 
-    init {
-        setOnDataLoadedCallback(onDataLoadedCallback)
-        searchStories = provideSearchStoriesUseCase(context)
-    }
+    private fun onDataLoaded(data: List<PlaidItem>) = onDataLoadedCallback.onDataLoaded(data)
```
ataulm: by inlining, I meant inlining this into the call sites; it's a single line function with the same signature (just delegated).

search/src/main/java/io/plaidapp/search/domain/SearchDataManager.kt
ataulm: I didn't get why not making it `val` - is this reassigned?

```kotlin
private var loadingCallbacks = mutableListOf<DataLoadingSubject.DataLoadingCallbacks>()
```

florina-muntenescu: I missed it here. updated

search/src/main/java/io/plaidapp/search/domain/SearchDataManager.kt
florina-muntenescu: for another PR that hopefully removes this completely

search/src/main/java/io/plaidapp/search/domain/SearchDataManager.kt
ataulm: perfick

PR #579, opened 2018-12-27
Adds a ViewModel to Search

florina-muntenescu:


florina-muntenescu:
`SearchDataManager` to be improved in another PR

core/src/main/java/io/plaidapp/core/designernews/domain/LoadStoriesUseCase.kt
```diff
@@ -65,7 +66,7 @@ class LoadStoriesUseCase @Inject constructor(
     }
 
     fun cancelAllRequests() {
-        parentJob.cancel()
+        parentJob.cancelChildren()
```
ataulm: I like this change - as far as this class is concerned, it doesn't know it's not going to be used again (looks like the call stack leads up to `HomeActivity.onDestroy` only); `cancelAllRequests()` is as-named 👌

search/src/main/java/io/plaidapp/search/dagger/SearchModule.kt
```diff
@@ -18,39 +18,55 @@ package io.plaidapp.search.dagger
...
-        DataManagerModule::class,
         DribbbleDataModule::class,
         FilterAdapterModule::class,
-        OnDataLoadedModule::class
+        SharedPreferencesModule::class
     ]
 )
-object SearchModule {
-    @JvmStatic
+class SearchModule(private val activity: SearchActivity) {
```
ataulm: Did this need to be changed back to a class?

None of the new methods below need `SearchActivity` as a type - the `searchViewModel` producer can have the `activity: Activity` as a parameter:

```kotlin
@Provides
fun searchViewModel(activity: Activity, factory: SearchViewModelFactory): SearchViewModel {
    return ViewModelProviders.of(activity, factory).get(SearchViewModel::class.java)
}
```

search/src/main/java/io/plaidapp/search/dagger/SearchModule.kt
```diff
@@ -18,39 +18,55 @@ package io.plaidapp.search.dagger
...
     @Provides
     fun isPocketInstalled(activity: Activity) = PocketUtils.isPocketInstalled(activity)
+
+    @Provides
+    fun searchViewModel(factory: SearchViewModelFactory): SearchViewModel {
+        return ViewModelProviders.of(activity, factory).get(SearchViewModel::class.java)
+    }
+
+    @Provides
+    fun searchViewModelFactory(dataManager: SearchDataManager): SearchViewModelFactory {
```
ataulm: If `SearchViewModelFactory` is only required to create a `SearchViewModel` - this function can be inlined in the `searchViewModel` producer function:

     @Provides
     fun searchViewModel(dataManager: SearchDataManager): SearchViewModel {
         val factory = SearchViewModelFactory(dataManager)
         return ViewModelProviders.of(activity, factory).get(SearchViewModel::class.java)
     }

search/src/main/java/io/plaidapp/search/domain/SearchDataManager.kt
```diff
@@ -36,10 +36,11 @@ import javax.inject.Inject
  */
 class SearchDataManager @Inject constructor(
     context: Context,
-    private val onDataLoadedCallback: OnDataLoadedCallback<List<PlaidItem>>,
     private val shotsRepository: ShotsRepository
 ) : DataLoadingSubject, LoadSourceCallback {
 
+    var onDataLoadedCallback: OnDataLoadedCallback<List<PlaidItem>>? = null
```
ataulm: this is never initialised as far as I can see. Is this related to:

>SearchDataManager to be improved in another PR

search/src/main/java/io/plaidapp/search/ui/SearchActivity.kt
```diff
@@ -244,7 +237,7 @@ class SearchActivity : Activity() {
     private fun doSave() {
         val saveData = Intent()
 
-        saveData.putExtra(Activities.Search.EXTRA_QUERY, dataManager.query)
+        saveData.putExtra(Activities.Search.EXTRA_QUERY, viewModel.getQuery())
```
ataulm: how come this bit was retrieved from the dataManager/viewModel but the others (e.g. `saveDribbble.isChecked`) are fetched from the View?

search/src/main/java/io/plaidapp/search/dagger/SearchModule.kt
ataulm: >My preference is to revert back to object.

what a terrible comment, given I provided no reasoning. I'm sure I understood it all when @hzsweers explained it in this pr https://github.com/nickbutcher/plaid/pull/539

core/src/main/java/io/plaidapp/core/designernews/domain/LoadStoriesUseCase.kt
florina-muntenescu: Actually, this change was made because once a job is cancelled it can no longer be active (see [here](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/-job/index.html)). Since the first thing we're doing with the use case is make sure we're canceling all requests, calling `cancel` lead to the coroutines never to be started. 
With [`cancelChildren`](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/cancel-children.html) the parent job is not affected.

search/src/main/java/io/plaidapp/search/dagger/SearchModule.kt
florina-muntenescu: I have no idea why you linked that PR. 
I'm using the `SearchActivity` here because `ViewModelProviders.of` method doesn't accept an `Activity` object ([doc](https://developer.android.com/reference/android/arch/lifecycle/ViewModelProviders)). So, I made `SearchActivity` be an `AppCompatActivity`, which extends from `FragmentActivity`, which is accepted by the `of` method.

search/src/main/java/io/plaidapp/search/domain/SearchDataManager.kt
florina-muntenescu: see the `init` block in `SearchViewModel`, where `onDataLoadedCallback` is set.

search/src/main/java/io/plaidapp/search/ui/SearchActivity.kt
florina-muntenescu: using the view now.

core/src/main/java/io/plaidapp/core/designernews/domain/LoadStoriesUseCase.kt
ataulm: :+1: I only checked for external usages of this method and didn't see it was called from inside here too.

search/src/main/java/io/plaidapp/search/dagger/SearchModule.kt
ataulm: I linked that PR because it was the one where Zac changed SearchModule to be an object.

If that's the only reason for the change, I'd recommend changing SearchComponentBuilder to accept SearchActivity or AppCompatActivity or FragmentActivity instead of Activity, then this class doesn't need changing to class and the activity doesn't need to be passed as a ctor parameter.

A change will be required to the producer parameters though from `activity: Activity` to `activity: SearchActivity` since Dagger won't understand that the bound instance can be casted to `Activity`.

hzsweers: I agree with Ataul on this. If you just make the type a SearchActivity or AppCompatActivity, you could keep this as an `object` and avoid having two-ish references of the instance on the graph.

Sidenote - the lack of scopes on any of these make it quite easy to accidentally end up with more than one instance of objects.

search/src/main/java/io/plaidapp/search/dagger/SearchModule.kt
```diff
@@ -18,39 +18,53 @@ package io.plaidapp.search.dagger
...
```
hzsweers: Actually I don't think you need this provider at all. Since `SearchViewModelFactory` has constructor injection, you can just just let dagger manage it for you (i.e. just add it as a parameter to `searchViewModel(...)` and dagger will wire the rest)

search/src/main/java/io/plaidapp/search/ui/SearchViewModel.kt
```diff
@@ -0,0 +1,70 @@
...
+    fun searchFor(query: String) {
+        dataManager.searchFor(query)
+    }
+
+    fun loadMore() {
+        dataManager.loadMore()
+    }
+
+    override fun onCleared() {
+        super.onCleared()
```
hzsweers: nit - super calls should conventionally go last in in destruction callbacks right?

search/src/main/java/io/plaidapp/search/ui/SearchViewModel.kt
```diff
@@ -0,0 +1,70 @@
...
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+import io.plaidapp.core.data.DataLoadingSubject
+import io.plaidapp.core.data.OnDataLoadedCallback
+import io.plaidapp.core.data.PlaidItem
+import io.plaidapp.core.util.event.Event
+import io.plaidapp.search.domain.SearchDataManager
+import javax.inject.Inject
+
+class SearchViewModel @Inject constructor(
```
hzsweers: is this `@Inject` used/necessary? I thought this could only be created by the ViewModelFactory

search/src/testDebug/java/io/plaidapp/search/TestData.kt
```diff
@@ -0,0 +1,57 @@
...
+
+import io.plaidapp.core.dribbble.data.api.model.Images
+import io.plaidapp.core.dribbble.data.api.model.Shot
+import io.plaidapp.core.dribbble.data.api.model.User
+
+/**
+ * Search test data
+ */
+
+val player1 = User(
```
hzsweers: This might make writing these easier - https://github.com/utwyko/KMinRandom

search/src/testDebug/java/io/plaidapp/search/ui/SearchViewModelTest.kt
```diff
@@ -0,0 +1,84 @@
...
+import org.junit.Rule
+import org.junit.Test
+
+/**
+ * Tests for [SearchViewModel] that mocks the dependencies
+ */
+class SearchViewModelTest {
+
+    // Executes tasks in the Architecture Components in the same thread
+    @get:Rule
```
hzsweers: nit: you can make this a `val` by annotating it as `@JvmField` instead. Or maybe you can make it a `val` regardless. In any case, the convention I've always seen/used for junit rules is to make them `@JvmField`

search/src/main/java/io/plaidapp/search/dagger/SearchModule.kt
```diff
@@ -67,4 +64,9 @@ class SearchModule(private val activity: SearchActivity) {
     fun searchViewModelFactory(dataManager: SearchDataManager): SearchViewModelFactory {
         return SearchViewModelFactory(dataManager)
     }
+
+    @Provides
+    fun dataLoadingSubject(searchDataManager: SearchDataManager): DataLoadingSubject {
```
hzsweers: If you make the module `abstract` you can just make this an abstract `@Binds` function

```kotlin
@Binds fun dataLoadingSubject(searchDataManager: SearchDataManager): DataLoadingSubject
```

search/src/main/java/io/plaidapp/search/ui/SearchViewModelFactory.kt
```diff
@@ -24,7 +24,6 @@ import javax.inject.Inject
 /**
  * Factory to create [SearchViewModel]
  */
-
```
hzsweers: Here I thought this doc was just trying to live dangerously

search/src/main/java/io/plaidapp/search/dagger/SearchModule.kt
florina-muntenescu: `@BindsInstance fun activity(activity: Activity)` is needed for the `FeedAdapter`, since that one needs a reference to an `android.app.Activity`. Any other ways to get rid of it?

ataulm: Yep, in the module, we need to tell Dagger that the same activity instance can be used for both.

If we make the change in the component to set `AppCompatActivity`, then in the module we can say:

```kotlin
@Provides
fun providesActivity(appCompatActivity: AppCompatActivity): Activity = appCompatActivity
```

florina-muntenescu: Ended up making a bigger change for this. Thanks @hzsweers for the help!

florina-muntenescu:
Updated. Ready for another check.

ataulm:
Thanks 👍 the [first part of this article](https://proandroiddev.com/dagger-2-annotations-binds-contributesandroidinjector-a09e6a57758f) helped me understand that binds is a shorthand (or has the effect of being a shorthand) for a producer function that returns the single parameter (I guess it wouldn't work if an `@Binds` annotated function had multiple parameters, or if it didn't have a return type).

it covers also that binds requires an abstract class (I guess it's generating the implementation) while provides functions must be static (hence moving the provides to a companion object). In a Kotlin object, I think you explained in that other pr I linked, that although functions in a Kotlin object are technically instance methods, the Java code that's generated will be static?

florina-muntenescu:
Thanks @ataulm and @hzsweers for reviews!

PR #582, opened 2018-12-31
Dagger consistency - use the constructor injection instead of the pro…

florina-muntenescu:
…vides methods


florina-muntenescu:
Thanks, @ataulm!

Issue #573, opened 2018-12-14
[Dagger] Add scopes to dagger components / modules


keyboardsurfer:
Currently there are no scopes added.
This means that we might get new instances where we would like to actually share them.

manuelvicnt:
This can be closed now. Both #649 and #662 fixes the issue.

PR #585, opened 2019-01-02
[Bug fix] Feed is not updated when changing filters

florina-muntenescu:
The proper fix for this would use `@Singleton` from dagger, but since we're only adding scopes later (see #573 ) I'm keeping this PR small, just to fix the bug.

### DIFF
--- a/core/src/main/java/io/plaidapp/core/dagger/FilterAdapterModule.kt
+++ b/core/src/main/java/io/plaidapp/core/dagger/FilterAdapterModule.kt
@@ -29,4 +29,10 @@ import io.plaidapp.core.ui.FilterAdapter
 @Module class FilterAdapterModule(val context: Context) {
 
     @Provides fun provideSources(): MutableList<Source> = SourceManager.getSources(context)
+
+    @Provides
+    fun provideFilterAdapter(
+        context: Context,
+        filters: MutableList<Source>
+    ) = FilterAdapter.getInstance(context, filters)
 }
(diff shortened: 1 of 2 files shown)


ataulm:
My feed (build from a couple days ago) does update when I toggle them (though not as expected).

florina-muntenescu:
> My feed (build from a couple days ago) does update when I toggle them (though not as expected).

With the master code it wasn't for me. You had to close the app and open the app again to see the feed from the right sources.

ataulm:
What was the bug? This fix makes it look like the filter adapter is recreated (and that it had state that wasn't restored, hence the singleton) whenever a filter is toggled. Is that the case?

florina-muntenescu:
> What was the bug? This fix makes it look like the filter adapter is recreated (and that it had state that wasn't restored, hence the singleton) whenever a filter is toggled. Was that the case?

Behaviour: when tapping on a filter, the feed doesn't update accordingly. For example, enabling DesignerNews, doesn't lead to DesignerNews items appearing
Root cause: `HomeActivity` and `DataManager` are using different instances of the `FilterAdapter`, therefore, only `HomeActivity` is notified when something changed. The fix uses only once instance of this. 
A proper-proper fix, would be removing the `FilterAdapter` completely from the `DataManager` (why would a DataManager know about a `RecyclerView.Adapter`????) but, until we get there, I want to make sure that at least we don't have this bug.

ataulm:
ahh got it, thank you 🙌

PR #586, opened 2019-01-02
Moves things around the FilterAdapter

florina-muntenescu:
* Classes out of FilterAdapter
* Functionality from the FilterAnimator
* Unused methods from FilterAdapter are removed


core/src/main/java/io/plaidapp/core/ui/filter/FilterAdapter.kt
```diff
@@ -0,0 +1,189 @@
...
+import io.plaidapp.core.ui.filter.FilterHolderInfo.Companion.FILTER_DISABLED
+import io.plaidapp.core.ui.filter.FilterHolderInfo.Companion.FILTER_ENABLED
+import io.plaidapp.core.ui.filter.FilterHolderInfo.Companion.HIGHLIGHT
+import io.plaidapp.core.ui.recyclerview.FilterSwipeDismissListener
+import java.util.Collections
+
+/**
+ * Adapter for showing the list of data sources used as filters for the home grid.
+ */
+class FilterAdapter(
```
ataulm: is this the thing in the right drawer? with the three items?

ataulm:
It looks like this  change is _just_ cutting up FilterAdapter into multiple files (+ removing the unused methods), is that right?

👍 lgtm if so, I don't want to add comments to change existing code if it's not relevant.

core/src/main/java/io/plaidapp/core/ui/filter/FilterAdapter.kt
florina-muntenescu: 3 items at the beginning. but once you search, you can also save the query and get it in your feed.

florina-muntenescu:
> It looks like this change is _just_ cutting up FilterAdapter into multiple files (+ removing the unused methods), is that right?
> 
> 👍 lgtm if so, I don't want to add comments to change existing code if it's not relevant.

Exactly! Just moving things around, to make it easier to refactor later

PR #587, opened 2019-01-02
[Bug fix] Save filter doesn't work.

florina-muntenescu:
Root cause: we were passing a CharSequence in the Intent extra and trying to read it as a String. This lead to getting null.


florina-muntenescu:
Thanks, @hzsweers!

PR #589, opened 2019-01-03
Creates a local data source for sources

florina-muntenescu:
Improves working with sources:
* Local data source created
* `SourceDataManager` renamed to `SourceRepository`
* `SourceRepository` receives now the defaults as a constructor param.


core/src/main/java/io/plaidapp/core/data/prefs/SourcesLocalDataSource.kt
```diff
@@ -0,0 +1,88 @@
...
+
+    /**
+     * Get all sources
+     */
+    fun getSources(): Set<String>? = getMutableSources()
+
+    /**
+     * Add a source and set the active state for the source
+     */
+    fun addSource(source: String, isActive: Boolean) {
```
ataulm: Could `isActive:Boolean` be more descriptive to the API as an enum?

```kotlin
enum class SourceState {
    ENABLED, DISABLED
}
```

When I was reading the tests, I couldn't tell what the booleans were doing.

core/src/main/java/io/plaidapp/core/data/prefs/SourcesLocalDataSource.kt
```diff
@@ -0,0 +1,88 @@
...
+
+/**
+ * Class that knows how to store locally sources, together with their active state
+ */
+class SourcesLocalDataSource(private val prefs: SharedPreferences) {
+
+    /**
+     * Get all sources
+     */
+    fun getSources(): Set<String>? = getMutableSources()
```
ataulm: What do these strings represent? In the test, I wasn't sure what 

```kotlin
dataSource.addSource(source, ...)
```

meant - is the word "source" being used to represent two different concepts?

core/src/main/java/io/plaidapp/core/dagger/SourcesRepositoryModule.kt
```diff
@@ -0,0 +1,76 @@
...
+import io.plaidapp.core.data.prefs.SourcesRepository
+
+/**
+ * Module to provide [SourcesRepository].
+ */
+@Module
+class SourcesRepositoryModule {
+
+    @Provides
+    fun provideSourcesLocalDataSource(context: Context): SourcesLocalDataSource {
```
ataulm: This could be inlined in the producer function for SourcesRepository unless it's needed as a dependency in other classes:

```kotlin
@Provides
fun provideSourceRepository(
    context: Context
): SourcesRepository {
    val defaultSources = provideDefaultSources(context)
    val sharedPrefs = context.getSharedPreferences(SOURCES_PREF, Context.MODE_PRIVATE)
    val localDataSource = SourcesLocalDataSource(sharedPrefs)
    return SourcesRepository(defaultSources, localDataSource)
}
```

or made similar to `provideDefaultSources` if you wanted to keep it as a separate function.

core/src/main/java/io/plaidapp/core/data/prefs/SourcesLocalDataSource.kt
ataulm: ah got it. For me, it would be helpful if this function was renamed to `getKeys()`.

This `LocalDataSource` isn't persisting `Source` types, it's storing keys - maybe the class could be renamed?

core/src/main/java/io/plaidapp/core/data/prefs/SourcesLocalDataSource.kt
florina-muntenescu: I would do that in the PR that changes the sources. To be honest, for me the entire `Source` class is confusing and can be nicely refactored. This PR already got quite big, to make this change here also.

core/src/main/java/io/plaidapp/core/ui/filter/FilterAdapter.kt
```diff
@@ -66,26 +71,26 @@ class FilterAdapter(
...
             }
         }
         // didn't already exist, so add it
-        filters.add(toAdd)
+        _filters.add(toAdd)
         Collections.sort(filters, Source.SourceComparator())
         dispatchFiltersChanged(toAdd)
         notifyDataSetChanged()
-        SourceManager.addSource(toAdd, context)
+        sourcesRepository.addSource(toAdd)
```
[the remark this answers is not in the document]
florina-muntenescu: Yes, exactly. `SourceRepository` should be the only source of truth for sources. Right know, that's pretty much `FilterAdapter`. The `FilterAdapter` should just get the list of sources and that's it. It shouldn't modify it. 
This is the first step in that direction.

ataulm:
>I don't fully get your PR comment

It's because I edited it while I was typing and deleted some sentences 🤦‍♂️ 

No. 4 is what I was driving at though.

>Understand the best way to model Plaid sources, in Kotlin. That also raises the question about whether these should be part of each module, not part of core

Thanks!

core/src/main/java/io/plaidapp/core/data/prefs/SourcesLocalDataSource.kt
florina-muntenescu: Keeping `SourceLocalDataSource` since we're saving keys and active states.
Renaming the method

core/src/main/java/io/plaidapp/core/data/prefs/SourcesLocalDataSource.kt
```diff
@@ -20,20 +20,20 @@ import android.content.SharedPreferences
...
- * Class that knows how to store locally sources, together with their active state
+ * Class that knows how to store locally sources keys, together with their active state
  */
 class SourcesLocalDataSource(private val prefs: SharedPreferences) {
 
     /**
      * Get all sources
      */
-    fun getSources(): Set<String>? = getMutableSources()
+    fun getKeys(): Set<String>? = getMutableKeys()
```
ataulm: 🎉

core/src/androidTest/java/io/plaidapp/core/data/prefs/SourcesLocalDataSourceTest.kt
```diff
@@ -45,13 +45,13 @@ class SourcesLocalDataSourceTest {
...
 
     @Test
-    fun getSources_returnsNullIfNothingWasAdded() {
+    fun getKeys_returnsNullIfNothingWasAdded() {
         // Given that we haven't saved any source
 
-        // When getting the sources
-        val sources = dataSource.getSources()
+        // When getting the keys
+        val sources = dataSource.getKeys()
```
ataulm: few instances of keys referred to as sources in this file - no big deal for me; change it if you touch this file next time?

florina-muntenescu: i promise!

Issue #588, opened 2019-01-03
found error: Cause: org.jetbrains.plugins.gradle.tooling.util.ModuleComponentIdentifierImpl.getModuleIdentifier()Lorg/gradle/api/artifacts/ModuleIdentifier;


poel:
When I import the project into android studio and compile, found error: 
Cause:org.jetbrains.plugins.gradle.tooling.util.ModuleComponentIdentifierImpl.getModuleIdentifier()Lorg/gradle/api/artifacts/ModuleIdentifier;

poel:
My android studio version is : 3.2.1

ataulm:
Hi @poel, thanks for the report.

I see the README still says that Plaid requires 3.2 or above, but I think that's outdated - it's using an alpha version of the Android Gradle Plugin which requires using AS 3.4 canaries if you want to build with AS (command-line builds should be fine).

I think they'd appreciate a PR to update the README if you learn which version specifically if necessary.

poel:
@ataulm Thanks for your feedback. I will update my AS to 3.4.

poel:
@ataulm I had update my AS to 3.4 and had run the project successfully! Thanks.

PR #590, opened 2019-01-03
Update README to reflect required AS version

ataulm:


ataulm:
Fixes https://github.com/nickbutcher/plaid/issues/588

PR #593, opened 2019-01-06
Cleaning up FilterAdapter and its usages

florina-muntenescu:
Finally the data layer doesn't have any more references to the UI layer! 🎉


app/src/main/java/io/plaidapp/ui/HomeActivity.java
```diff
@@ -119,16 +120,17 @@
...
     private boolean monitoringConnectivity = false;
+    private FilterAdapter filtersAdapter;
 
     // data
     @Inject
     DataManager dataManager;
     @Inject
     FeedAdapter adapter;
     @Inject
-    FilterAdapter filtersAdapter;
```
ataulm: 🙌

core/src/main/java/io/plaidapp/core/data/prefs/SourcesRepository.kt
```diff
@@ -27,6 +28,12 @@ class SourcesRepository(
     private val dataSource: SourcesLocalDataSource
 ) {
 
+    private val callbacks = mutableListOf<FiltersChangedCallback>()
+
+    fun registerFilterChangedCallback(callback: FiltersChangedCallback) {
```
ataulm: is this pattern expected to stay here long term?  I would expect eventually that `getSources()` would return `LiveData<List<Source>>`/`Observable<List<Source>>`.

core/src/main/java/io/plaidapp/core/ui/filter/FilterAdapter.kt
```diff
@@ -35,18 +35,14 @@ class FilterAdapter(
     private val sourcesRepository: SourcesRepository
 ) : RecyclerView.Adapter<FilterViewHolder>(), FilterSwipeDismissListener {
 
-    private var callbacks: MutableList<FiltersChangedCallback> = ArrayList()
-
     val enabledSourcesCount: Int
         get() = filters.count(Source::active)
 
-    private val _filters: MutableList<Source>
-    val filters: List<Source>
```
ataulm: Was this unused already? I can't see usages of `.filters` in this PR.

core/src/main/java/io/plaidapp/core/data/prefs/SourcesRepository.kt
florina-muntenescu: yeah, we're definitely going to change it to something better than callbacks. we haven't decided on what for now.

core/src/main/java/io/plaidapp/core/ui/filter/FilterAdapter.kt
florina-muntenescu: it was used from `DataManger`, for example, which is still a java file, so it was `.getFilters`

ataulm: ♂️oh yea

PR #594, opened 2019-01-06
Moves enabled sources count to SourcesRepository

florina-muntenescu:


core/src/main/java/io/plaidapp/core/data/prefs/SourcesRepository.kt
```diff
@@ -86,6 +86,10 @@ class SourcesRepository(
         dispatchSourceRemoved(source)
     }
 
+    fun getActiveSourcesCount(): Int {
+        return getSources().filter { it.active }.count()
```
ataulm: also `return getSources().count { it.active }`

florina-muntenescu: I knew there had to be something better! Thanks

ataulm: I've been burned too many times so I checked - there's always something better 😅

PR #595, opened 2019-01-06
Adds an in-memory cache to SourcesRepository

florina-muntenescu:


core/src/main/java/io/plaidapp/core/data/prefs/SourcesRepository.kt
```diff
@@ -73,16 +79,22 @@ class SourcesRepository(
...
+        cache.find { it.key == source.key }?.apply { active = source.active }
         dispatchSourceChanged(source)
     }
 
     fun removeSource(source: Source) {
         dataSource.removeSource(source.key)
+        val index = cache.indexOf(source)
+        if (index >= 0) {
+            cache.removeAt(index)
+        }
```
ataulm: `cache.remove(source)` should be enough too?

core/src/main/java/io/plaidapp/core/data/prefs/SourcesRepository.kt
```diff
@@ -28,13 +28,18 @@ class SourcesRepository(
...
 ) {
 
+    private val cache = mutableListOf<Source>()
     private val callbacks = mutableListOf<FiltersChangedCallback>()
 
     fun registerFilterChangedCallback(callback: FiltersChangedCallback) {
         callbacks.add(callback)
     }
 
     fun getSources(): List<Source> {
```
ataulm: 1) is it ok that we're returning the cache instance directly?
2) this fun could be split into two now:

```kotlin
fun getSources(): List<Source> {
    if (cache.isEmpty()) {
        val sources = getSourcesFromSomethingBetterFunctionName()
        cache.addAll(sources)
    }
    return cache.toList()
}

private fun getSourcesFromSomethingBetterFunctionName(): List<Source> {
    // the old function
}

```

florina-muntenescu: But I am returning it as a `List<Source>`, so when it's called from the outside, you don't have access to the `MutableList` methods. You'd need to specifically cast ` repository.getSources() as MutableList` to use it as such. 
What would be the benefit of `cache.toList()`?

ataulm: there isn't much - it'll be very defensive.

florina-muntenescu: I think it's too much. As long as we're returning it as `List`, it should be ok.

PR #596, opened 2019-01-06
Improving HomeActivity

florina-muntenescu:


app/src/main/java/io/plaidapp/dagger/HomeModule.kt
```diff
@@ -28,32 +31,54 @@ import io.plaidapp.core.dagger.SourcesRepositoryModule
...
+            OnDataLoadedModule::class
+        ]
 )
-class HomeModule(private val activity: Activity) {
+abstract class HomeModule {
 
-    @Provides
-    fun context(): Context = activity
+    @Binds
+    abstract fun homeActivityAsFragmentActivity(activity: HomeActivity): FragmentActivity
```
hzsweers: Random but if you find you use this pattern a lot, I think you can pull out these FragmentActivity/Activity/Context bindings into an interface or base abstract class (CC @ronshapiro)

florina-muntenescu: CC @keyboardsurfer

ronshapiro: Agreed. the `@BindsInstance` method can just take the type that you actually need.

Ideally it would be good to converge on one type for the bindings - `FragmentActivity` is probably the best lowest common denominator.

I'm also always wary of binding `Context`, because there are so many `Context`s out there.

florina-muntenescu:
The Home Dagger related changes have been copied to #623 to decrease the scope of this PR.

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -0,0 +1,311 @@
...
+        Mockito.verify(sourcesRepository).registerFilterChangedCallback(
+            capture(filtersChangedCallback)
+        )
+        // Given that filters were updated
+        filtersChangedCallback.value.onFiltersUpdated(listOf(designerNewsSource))
+        // Given that ui model sources are emitted
+        val sources = LiveDataTestUtil.getValue(homeViewModel.sources)
+        val uiSource = sources!!.sourceUiModels[0]
+
+        // When calling onSourceRemoved
```
nickbutcher: onSourceDismissed

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -0,0 +1,311 @@
...
```
nickbutcher: onSourceDismissed

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -0,0 +1,311 @@
...
+        // Given a view model
+        val homeViewModel = createViewModelWithDefaultSources(emptyList())
+        Mockito.verify(sourcesRepository).registerFilterChangedCallback(
+            capture(filtersChangedCallback)
+        )
+        // Given that filters were updated
+        filtersChangedCallback.value.onFiltersUpdated(listOf(designerNewsSource))
+        // Given that ui model sources are emitted
+        val sources = LiveDataTestUtil.getValue(homeViewModel.sources)
+        val uiSource = sources!!.sourceUiModels[0]
```
nickbutcher: What about `!!` the previous line? I know we only access `sources` once but conceptually that feels like the right place.

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -0,0 +1,311 @@
...
+        Mockito.verify(sourcesRepository).registerFilterChangedCallback(
+            capture(filtersChangedCallback)
+        )
+
+        // When an inactive source was changed
+        val activeSource = Source.DribbbleSearchSource("dribbble", true)
+        filtersChangedCallback.value.onFiltersChanged(activeSource)
+
+        // Then source removed value is null
+        val source = LiveDataTestUtil.getValue(homeViewModel.sourceRemoved)
```
nickbutcher: Is this the expected behaviour? Why do we emit null here?

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -0,0 +1,311 @@
...
+    @Test
+    fun filtersChanged_activeSource() {
+        // Given a view model
+        val homeViewModel = createViewModelWithDefaultSources(emptyList())
+        Mockito.verify(sourcesRepository).registerFilterChangedCallback(
+            capture(filtersChangedCallback)
+        )
+
+        // When an inactive source was changed
+        val activeSource = Source.DribbbleSearchSource("dribbble", true)
```
nickbutcher: named args on active param for readability?

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -0,0 +1,311 @@
...
+            capture(filtersChangedCallback)
+        )
+
+        // When an inactive source was changed
+        val inactiveSource = Source.DribbbleSearchSource("dribbble", false)
+        filtersChangedCallback.value.onFiltersChanged(inactiveSource)
+
+        // Then the source removed contains the inactive source
+        val source = LiveDataTestUtil.getValue(homeViewModel.sourceRemoved)
+        assertEquals(inactiveSource, source)
```
nickbutcher: Again, not sure why changing active state of a source is emitting the removed event?

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -0,0 +1,311 @@
...
+    }
+
+    private fun createViewModelWithDefaultSources(list: List<Source>): HomeViewModel = runBlocking {
```
nickbutcher: Default `list` to `emptyList` to clean up common usage?

nickbutcher: Not sure what `DefaultSources` refers to?

core/src/main/java/io/plaidapp/core/data/Source.java
```diff
@@ -46,6 +47,40 @@ public Source(String key,
...
 
+    public void setActive(boolean active) {
+        this.active = active;
+    }
+
+    public boolean getActive() {
+        return active;
+    }
+
+    @Override
```
nickbutcher: ⌥⇧⌘k

core/src/main/java/io/plaidapp/core/data/prefs/SourcesRepository.kt
```diff
@@ -35,7 +38,12 @@ class SourcesRepository(
         callbacks.add(callback)
     }
 
-    fun getSources(): List<Source> {
+    suspend fun getSources() = withContext(dispatcherProvider.io) {
+        getSourcesSync()
+    }
+
+    @Deprecated("Use the suspending getSourcesSync")
```
nickbutcher:
```suggestion
    @Deprecated("Use the suspending getSources")
```

core/src/main/java/io/plaidapp/core/data/prefs/SourcesRepository.kt
```diff
@@ -69,41 +77,73 @@ class SourcesRepository(
...
+        }
+        // they didn't already exist, so add them
+        addSources(sourcesToAdd)
     }
 
-    fun updateSource(source: Source) {
-        dataSource.updateSource(source.key, source.active)
-        cache.find { it.key == source.key }?.apply { active = source.active }
-        dispatchSourceChanged(source)
+    fun changeSourceActiveState(sourceKey: String, newActiveState: Boolean) {
```
nickbutcher: Shouldn't this be private? This last param seems strange and a smell. Do we only pass `false` for tests? Can we avoid exposing this?

core/src/main/java/io/plaidapp/core/data/prefs/SourcesRepository.kt
```diff
@@ -69,41 +77,73 @@ class SourcesRepository(
...
+                    sourcePresent = true
+                    // already exists, just ensure it's active
+                    if (!existing.active) {
+                        changeSourceActiveState(existing.key, true)
+                    }
+                }
+            }
+            if (!sourcePresent) {
+                // doesn't exist so needs to be added
+                sourcesToAdd.add(toAdd)
```
nickbutcher: `+=`

core/src/main/java/io/plaidapp/core/data/prefs/SourcesRepository.kt
```diff
@@ -69,41 +77,73 @@ class SourcesRepository(
...
+        dispatchSourcesUpdated()
     }
 
-    fun removeSource(source: Source) {
-        dataSource.removeSource(source.key)
-        cache.remove(source)
-        dispatchSourceRemoved(source)
+    fun removeSource(sourceKey: String) {
+        dataSource.removeSource(sourceKey)
+        cache.find { it.key == sourceKey }?.let {
```
nickbutcher: could this just be `if (cache.remove(sourceKey)) { …`

core/src/main/java/io/plaidapp/core/ui/filter/FilterAdapter.kt
```diff
@@ -135,22 +79,37 @@ class FilterAdapter(
...
-        }
+        val uiModel = getItem(position)
+        uiModel.onSourceDismissed(uiModel)
     }
 
     companion object {
 
-        internal const val FILTER_ICON_ENABLED_ALPHA = 179 // 70%
-        internal const val FILTER_ICON_DISABLED_ALPHA = 51 // 20%
+        private val sourceUiModelDiff = object : DiffUtil.ItemCallback<SourceUiModel>() {
```
nickbutcher: Does this need to be in the Companion? Isn't being an object enough to make it a singleton?

core/src/main/java/io/plaidapp/core/ui/filter/FilterViewHolder.kt
```diff
@@ -104,4 +103,9 @@ class FilterViewHolder(itemView: View) : RecyclerView.ViewHolder(itemView) {
             }
         }
     }
+
+    companion object {
```
nickbutcher: Maybe make these private top level constants as we're not using outside of this scope?

core/src/main/java/io/plaidapp/core/ui/filter/SourceUiModel.kt
```diff
@@ -0,0 +1,51 @@
...
+ * UI model for a source
+ */
+data class SourceUiModel(
+    val key: String,
+    val name: String,
+    val active: Boolean,
+    @DrawableRes val iconRes: Int,
+    val isSwipeDismissable: Boolean,
+    val onSourceClicked: (source: SourceUiModel) -> Unit,
+    val onSourceDismissed: (source: SourceUiModel) -> Unit
```
nickbutcher: Should this be nullable for non swipe-dismissable sources?

app/src/main/java/io/plaidapp/ui/HomeViewModel.kt
```diff
@@ -0,0 +1,189 @@
...
+
+/**
+ * ViewModel for [HomeActivity].
+ * Handles user login status and sources display for the drawer.
+ *
+ * TODO: handling of the feed list
+ */
+class HomeViewModel(
+    val dataManager: DataManager,
+    private val loginRepository: LoginRepository,
```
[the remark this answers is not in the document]
florina-muntenescu: the class is not called `LoginRepository` and we don't have any other login in the app. Why do you think this should be renamed?

app/src/main/java/io/plaidapp/ui/HomeViewModel.kt
```diff
@@ -0,0 +1,189 @@
...
+        get() = _sourceRemoved
+
+    private val _sources = MutableLiveData<SourcesUiModel>()
+    val sources: LiveData<SourcesUiModel>
+        get() = _sources
+
+    // listener for notifying adapter when data sources are deactivated
+    private val filtersChangedCallbacks = object : FiltersChangedCallback() {
+        override fun onFiltersChanged(changedFilter: Source) {
+            if (!changedFilter.active) {
```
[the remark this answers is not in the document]
florina-muntenescu: Deactivating a source is not removing it, that's correct. I added already a TODO above the `sourceRemoved` LiveData saying that this object is added temporarily until we deal with FeedAdapter. Right now, the behaviour when deactivating a source or removing a source is the same: we're removing feed items that belong to the deactivated/removed source.

app/src/main/java/io/plaidapp/ui/HomeViewModel.kt
```diff
@@ -0,0 +1,189 @@
...
+    fun loadData() {
+        dataManager.loadAllDataSources()
+    }
+
+    override fun onCleared() {
+        dataManager.cancelLoading()
+        super.onCleared()
+    }
+
+    fun addSources(query: String, isDribbble: Boolean, isDesignerNews: Boolean) {
```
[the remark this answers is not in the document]
florina-muntenescu: But if `isDribbble` is true and `isDesignerNews` is true, then I'm adding 2 sources.

app/src/main/java/io/plaidapp/ui/HomeViewModelFactory.kt
```diff
@@ -0,0 +1,45 @@
...
+import io.plaidapp.core.data.prefs.SourcesRepository
+import io.plaidapp.core.designernews.data.login.LoginRepository
+import javax.inject.Inject
+
+/**
+ * Factory to create [HomeViewModel]
+ */
+class HomeViewModelFactory @Inject constructor(
+    private val dataManager: DataManager,
+    private val loginRepository: LoginRepository,
```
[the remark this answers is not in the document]
florina-muntenescu: same comment as above

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -0,0 +1,311 @@
...
+        // Given a login status
+        whenever(loginRepository.isLoggedIn).thenReturn(false)
+
+        // When getting the login status
+        val isLoggedIn = homeViewModel.isDesignerNewsUserLoggedIn()
+
+        // The login status is the expected one
+        assertFalse(isLoggedIn)
+    }
+
```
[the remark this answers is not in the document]
florina-muntenescu: I changed the test to pass a variable for log in status.

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -0,0 +1,311 @@
...
+        homeViewModel.addSources("query", isDribbble = true, isDesignerNews = true)
+
+        // Then two sources are added to the repository
+        val expected = listOf(
+            Source.DribbbleSearchSource("query", true),
+            Source.DesignerNewsSearchSource("query", true)
+        )
+        verify(sourcesRepository).addOrMarkActiveSources(expected)
+    }
+
```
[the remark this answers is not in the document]
florina-muntenescu: I thought this is tested in `addSources_Dribbble`. What test are you expecting?

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -0,0 +1,311 @@
...
+        )
+
+        // When updating the filters to a new list of sources
+        filtersChangedCallback.value.onFiltersUpdated(listOf(designerNewsSource, dribbbleSource))
+
+        // Then ui model sources are emitted
+        val sources = LiveDataTestUtil.getValue(homeViewModel.sources)
+        // Then all sources are highlighted
+        val sourcesHighlightUiModel = SourcesHighlightUiModel(listOf(0, 1), 1)
+        assertEquals(Event(sourcesHighlightUiModel), sources?.highlightSources)
```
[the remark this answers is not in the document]
florina-muntenescu: By wrapping the expected in an Event, we also compare the `consumed` value.

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
florina-muntenescu: named args can't be used for java classes

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
florina-muntenescu: if the source changed to active, then we don't need to remove it. 
The `when` comment was incorrect and I assume that was misleading. Fixed the comment.

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
florina-muntenescu: See [comment](https://github.com/nickbutcher/plaid/pull/596/files#r262853301)

core/src/main/java/io/plaidapp/core/data/Source.java
florina-muntenescu: I don't think this is the time to do that. 
Converting this file to Kotlin won't help much, given that Source can't be a data class. 
I believe that the sources hierarchy can be improved but the consequences of that would highly increase the complexity of this PR. 
Suggesting that this file is converted to Kotlin in another PR

core/src/main/java/io/plaidapp/core/ui/filter/FilterAdapter.kt
florina-muntenescu: Made it a top level property 
I need to pass an instance to the constructor

app/src/main/java/io/plaidapp/ui/HomeViewModel.kt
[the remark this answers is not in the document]
nickbutcher: I think it makes it's responsibilities clearer. Currently you have to know that we don't have other login functionality within the app as the repo is named generically.

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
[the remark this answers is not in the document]
nickbutcher: Was that there before?! Can't tell with the force pushing. Nevermind.

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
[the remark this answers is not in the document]
nickbutcher: Isn't this the entire point of the `peek` method? Should we remove it if we want to make this change in how we're testing `Event`s?

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
nickbutcher: ewwwww Java

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -0,0 +1,318 @@
...
+        homeViewModel.logoutFromDesignerNews()
+
+        // Then logout is called
+        verify(loginRepository).logout()
+    }
+
+    @Test
+    fun isDesignerNewsLoggedIn() {
+        // Given a view model
+        val homeViewModel = createViewModel(emptyList())
```
nickbutcher: Use default param?

nickbutcher:
A few minor things but mostly looking good. Thanks for the changes :D

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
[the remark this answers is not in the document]
florina-muntenescu: using `peek`

app/src/main/java/io/plaidapp/ui/HomeViewModelFactory.kt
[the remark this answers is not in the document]
nickbutcher: change here also per above?

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -0,0 +1,317 @@
...
+        true
+    )
+    private val designerNewsSourceUiModel = SourceUiModel(
+        designerNewsSource.key,
+        designerNewsSource.name,
+        designerNewsSource.active,
+        designerNewsSource.iconRes,
+        designerNewsSource.isSwipeDismissable,
+        {},
+        {}
```
nickbutcher: Could now use `null` here?

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -0,0 +1,317 @@
...
+ * Tests for [HomeViewModel], with dependencies mocked.
+ */
+class HomeViewModelTest {
+
+    // Executes tasks in the Architecture Components in the same thread
+    @get:Rule
+    var instantTaskExecutorRule = InstantTaskExecutorRule()
+
+    private val designerNewsSource = Source.DesignerNewsSearchSource(
+        "query",
```
nickbutcher: named params for readability?

core/src/main/java/io/plaidapp/core/data/prefs/SourcesRepository.kt
```diff
@@ -69,47 +77,78 @@ class SourcesRepository(
...
+        // they didn't already exist, so add them
+        addSources(sourcesToAdd)
     }
 
-    fun updateSource(source: Source) {
-        dataSource.updateSource(source.key, source.active)
-        cache.find { it.key == source.key }?.apply { active = source.active }
-        dispatchSourceChanged(source)
+    fun changeSourceActiveState(sourceKey: String) {
+        cache.find { it.key == sourceKey }?.apply {
```
nickbutcher: `?.let` feels more idiomatic as it's not configuring the source.

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
florina-muntenescu: would love to but...java

Issue #612, opened 2019-01-28
Use gradle repo include groups


nickbutcher:
See https://jebware.com/blog/?p=573

rakeshgangwar:
Have opened a pull request for this issue.
https://github.com/nickbutcher/plaid/pull/615

florina-muntenescu:
Fixed by @rakeshgangwar. Thank you!

PR #615, opened 2019-02-18
Gradle includeGroup

rakeshgangwar:
Working on this up for grabs issue.
https://github.com/nickbutcher/plaid/issues/612

This is my first PR and contribution towards an open source project, please be soft on me.
Looking forward to reviews on this.


about/build.gradle
```diff
@@ -40,10 +40,64 @@ android {
 }
 
 repositories {
-    google()
-    mavenCentral()
```
keyboardsurfer: You removed mavenCentral completely. Is there a rationale behind that?

about/build.gradle
```diff
@@ -40,10 +40,64 @@ android {
...
+    maven {
+        url "http://maven.google.com"
+        content {
+            includeGroupByRegex "com\\.android.*"
+            includeGroupByRegex "androidx.*"
+            includeGroup "android.arch.lifecycle"
+            includeGroup "android.arch.core"
+            includeGroup "com.google.android.material"
+        }
+    }
```
keyboardsurfer: nit: new line after closing `maven {...}` makes it more legible

about/build.gradle
```diff
@@ -40,10 +40,64 @@ android {
...
+            includeGroup "org.glassfish.jaxb"
+            includeGroup "org.hamcrest"
+            includeGroup "org.jvnet.staxex"
+            includeGroup "org.jsoup"
+            includeGroup "org.mockito"
+            includeGroup "org.objenesis"
+            includeGroup "org.ow2.asm"
+            includeGroup "org.sonatype.oss"
+
+            excludeGroup "com.google.android.material"
```
keyboardsurfer: Is the explicit exclusion necessary?

about/build.gradle
```diff
@@ -40,10 +40,64 @@ android {
...
+            includeGroupByRegex "androidx.*"
+            includeGroup "android.arch.lifecycle"
+            includeGroup "android.arch.core"
+            includeGroup "com.google.android.material"
+        }
+    }
+    maven {
+        url 'https://jcenter.bintray.com'
+        content {
+            includeGroupByRegex "com\\.google.*"
```
keyboardsurfer: It feels like these should come from maven.google.com

about/build.gradle
```diff
@@ -40,10 +40,64 @@ android {
...
+            includeGroup "org.jsoup"
+            includeGroup "org.mockito"
+            includeGroup "org.objenesis"
+            includeGroup "org.ow2.asm"
+            includeGroup "org.sonatype.oss"
+
+            excludeGroup "com.google.android.material"
+        }
+    }
+    maven {
```
keyboardsurfer: Fabric is also available on maven.google.com. Can you switch it over to this?

about/build.gradle
rakeshgangwar: Various important com.google.* libraries are not present in maven.google.com. One of them being com.google.code.gson that is why I have added com.google.* in jcenter.

about/build.gradle
rakeshgangwar: I have added includeGroupByRegex for com.google.* and only this particular library comes from maven.google.com. That is why I have excluded it here so that gradle doesn't try to download this library from multiple repos.

about/build.gradle
rakeshgangwar: Initially I skipped jcenter completely by mistake and tried to download all the libraries from mavenCentral. Later I realized that several required libraries are not present in mavenCentral but in jCenter and then I added all mavenCentral libraries to jCenter and project passed all the tests without any issues.

In further reading about it, I came across this answer on stackoverflow https://stackoverflow.com/a/28457914, which says jCenter is a superset of mavenCentral.

about/build.gradle
rakeshgangwar: I will switch to google repository for this and update the pull request if it works.

rakeshgangwar:
@keyboardsurfer Thanks for the review. I have added replies on your comments.

Will look into the possibility of using `apply from:` and will update the PR accordingly. 👍

about/build.gradle
rakeshgangwar: Current fabric version we are using is v1.25.4 whereas on google repository v1.3.17 onwards are available. 
So, I think for now we can stick to maven.fabric.io unless we are planning to bump up the fabric version soon.

about/build.gradle
rakeshgangwar: I will add new lines and update the PR.

build.gradle
```diff
@@ -15,6 +15,7 @@
  */
 
 buildscript {
+    scriptHandler -> apply from: 'core_repositories.gradle', to: scriptHandler
```
keyboardsurfer: Didn't know about that one; Neat!

core_repositories.gradle
```diff
@@ -0,0 +1,70 @@
...
+
+    maven {
+        url 'https://maven.fabric.io/public'
+        content {
+            includeGroupByRegex "io\\.fabric.*"
+            includeGroupByRegex "com\\.crashlytics.*"
+        }
+    }
+
+    maven {
```
keyboardsurfer: Both crashlytics and fabric are on `maven.google.com`
Can we remove this closure in favor of m.g.c.?

third_party/bypass/build.gradle
```diff
@@ -18,9 +18,54 @@ android {
 }
 
 repositories {
```
keyboardsurfer: Does it make sense to `apply from` here as well?

rakeshgangwar: Yep. I must have missed it. Updating PR with `apply from` here as well.

core_repositories.gradle
rakeshgangwar: As mentioned in my previous comment current version being used of fabric v1.25.4 is not available on google repository. We can remove fabric repository once we upgrade the fabric to the latest version.

keyboardsurfer: SG, filed https://github.com/nickbutcher/plaid/issues/616 to track this.

rakeshgangwar:
@keyboardsurfer I have updated gradle version from beta02 to beta04 since it was failing to build on my machine.
After updating it was running successfully and has passed all the tests. Let me know if it creates any issue.

build.gradle
```diff
@@ -17,7 +17,7 @@
 buildscript {
     scriptHandler -> apply from: 'repositories.gradle', to: scriptHandler
 
-    ext.versions = [
+        ext.versions = [
```
keyboardsurfer: Nit: Please revert this indentation.

rakeshgangwar: Fixed.

rakeshgangwar:
@keyboardsurfer I had just made an indentation fix but tests are failing on BuildCI with `java.io.IOException: Cannot allocate memory`
Can you please tell me why this could be happening?

PR #617, opened 2019-02-20
Defer initialization of OkHttpClient until first request

ZacSweers:
This uses a pattern of deferred initialization via Retrofit's `Call.Factory` support to allow OkHttpClient (and its initialization) to be deferred until the first network request, and subsequently also move it entirely to a background thread.

I opportunistically cleaned up a couple things here too to either make it more idiomatic DI (intermediate dependencies), and also leverage use of private qualifiers to newBuilder an upstream OkHttpClient as needed rather than keep adding to a builder (which would have added those interceptors to other modules too since the builder's mutable!). This also shared the OkHttpClient instance with the dribbble module, which was otherwise just creating its own instance and doubling the connection pools.

### DIFF
--- a/core/src/main/java/io/plaidapp/core/dagger/CoreDataModule.kt
+++ b/core/src/main/java/io/plaidapp/core/dagger/CoreDataModule.kt
@@ -32,8 +32,8 @@ import retrofit2.converter.gson.GsonConverterFactory
 class CoreDataModule {
 
     @Provides
-    fun provideOkHttpClientBuilder(interceptor: HttpLoggingInterceptor): OkHttpClient.Builder =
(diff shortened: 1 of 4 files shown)


ZacSweers:
d'oh, the RUNTIME retention (kotlin equivalent of CLASS retention) is still needed to register them correctly across module boundaries because dagger still validates the qualified ones against the whole graph, even though _we_ know they're effectively private. FYI @ronshapiro, could maybe be an optimization case if we can mark some dependencies as known to be not accessible elsewhere?

ronshapiro:
We'd still need to serialize that somehow, and I feel like extra codegen isn't a good option when there's already a builtin mechanism in bytecode

PR #619, opened 2019-02-23
Fix false positive unit tests

arjunpola:
Fixes an issue where the assertion exception is
thrown in child coroutine scope instead of parent
scope when running tests in SearchStoriesUseCaseTest.kt

### DIFF
--- a/core/src/main/java/io/plaidapp/core/designernews/domain/SearchStoriesUseCase.kt
+++ b/core/src/main/java/io/plaidapp/core/designernews/domain/SearchStoriesUseCase.kt
@@ -22,6 +22,7 @@ import io.plaidapp.core.data.Result
 import io.plaidapp.core.designernews.data.stories.StoriesRepository
 import io.plaidapp.core.designernews.data.stories.model.toStory
 import kotlinx.coroutines.CoroutineScope
+import kotlinx.coroutines.CoroutineExceptionHandler
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.cancelChildren
 import kotlinx.coroutines.launch
@@ -37,12 +38,16 @@ class SearchStoriesUseCase @Inject constructor(
 ) {
     private var parentJob = Job()
     private val scope = CoroutineScope(dispatcherProvider.main + parentJob)
+    private val exceptionHandler: CoroutineExceptionHandler = CoroutineExceptionHandler { _, e ->
+        println("SearchStoriesUseCase:Exception - $e")
+    }
 
     private val parentJobs = mutableMapOf<String, Job>()
 
-    operator fun invoke(query: String, page: Int, callback: LoadSourceCallback) {
+    operator fun invoke(query: String, page: Int, callback: LoadSourceCallback): Job? {
         val jobId = "$query::$page"
         parentJobs[jobId] = launchRequest(query, page, callback, jobId)
+        return parentJobs[jobId]
     }
 
     private fun launchRequest(
@@ -50,16 +55,16 @@ class SearchStoriesUseCase @Inject constructor(
         page: Int,
         callback: LoadSourceCallback,
         jobId: String
-    ) = scope.launch(dispatcherProvider.computation) {
+    ) = scope.launch(dispatcherProvider.computation + exceptionHandler) {
         val result = storiesRepository.search(query, page)
         parentJobs.remove(jobId)
         if (result is Result.Success) {
             val stories = result.data.map { it.toStory() }
-            withContext(dispatcherProvider.main) {
+            withContext(dispatcherProvider.main + exceptionHandler) {
                 callback.sourceLoaded(stories, page, query)
             }
         } else {
-            withContext(dispatcherProvider.main) { callback.loadFailed(query) }
+            withContext(dispatcherProvider.main + exceptionHandler) { callback.loadFailed(query) }
         }
     }
 
--- a/core/src/main/java/io/plaidapp/core/designernews/domain/SearchStoriesUseCase.kt
+++ b/core/src/main/java/io/plaidapp/core/designernews/domain/SearchStoriesUseCase.kt
@@ -22,7 +22,6 @@ import io.plaidapp.core.data.Result
 import io.plaidapp.core.designernews.data.stories.StoriesRepository
 import io.plaidapp.core.designernews.data.stories.model.toStory
 import kotlinx.coroutines.CoroutineScope
-import kotlinx.coroutines.CoroutineExceptionHandler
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/plaidapp/core/designernews/domain/SearchStoriesUseCase.kt
```diff
@@ -37,29 +38,33 @@ class SearchStoriesUseCase @Inject constructor(
 ) {
     private var parentJob = Job()
     private val scope = CoroutineScope(dispatcherProvider.main + parentJob)
+    private val exceptionHandler: CoroutineExceptionHandler = CoroutineExceptionHandler { _, e ->
```
ataulm: Is it necessary to include a `CoroutineExceptionHandler` in the scope for the Job to be cancelled when an exception is thrown? If not, I think this can be removed.

core/src/main/java/io/plaidapp/core/designernews/domain/SearchStoriesUseCase.kt
```diff
@@ -37,29 +38,33 @@ class SearchStoriesUseCase @Inject constructor(
...
+        println("SearchStoriesUseCase:Exception - $e")
+    }
 
     private val parentJobs = mutableMapOf<String, Job>()
 
-    operator fun invoke(query: String, page: Int, callback: LoadSourceCallback) {
+    operator fun invoke(query: String, page: Int, callback: LoadSourceCallback): Job? {
```
ataulm: `LoadSourceCallback` is used to abstract the concept of coroutines from the use case. Adding `Job?` as a return type exposes this concept to clients of this class.

I'm not sure adding this just for the test is worth it.

core/src/test/java/io/plaidapp/core/designernews/domain/SearchStoriesUseCaseTest.kt
```diff
@@ -39,6 +40,7 @@ import java.util.GregorianCalendar
 /**
  * Tests for [SearchStoriesUseCase] mocking the dependencies.
  */
+@InternalCoroutinesApi
```
ataulm: [This annotation shouldn't be used outside of the `kotlinx.coroutines`](https://github.com/Kotlin/kotlinx.coroutines/blob/master/common/kotlinx-coroutines-core-common/src/Annotations.kt#L30-L32) - we should avoid using `getCancellationException()` since it's not part of the (intended) public (use) API.

core/src/main/java/io/plaidapp/core/designernews/domain/SearchStoriesUseCase.kt
arjunpola: No not necessary, added it to prevent the exception from being thrown twice. But, it would consume other exceptions if we do not check for them using the job.

Will remove it.

core/src/test/java/io/plaidapp/core/designernews/domain/SearchStoriesUseCaseTest.kt
arjunpola: Will remove. This was needed for getCancelableException. But, will look for an alternative way of fixing this.

core/src/main/java/io/plaidapp/core/designernews/domain/SearchStoriesUseCase.kt
arjunpola: Sure will remove this too.

arjunpola:
Thank you @ataulm for reviews and feedback :)

I also tried another approach of plumbing an optional coroutine context along with dispatchers. Using the context when provided can make sure that exceptions are propogated to parent scope.

Also, I like the idea of having a fake implementation of LoadSourceCallback, will explore that too 👍

arjunpola:
Realized the mistake in this approach. I have another patch with the callback approach.

core/src/test/java/io/plaidapp/core/designernews/domain/AwesomeTestLoadSourceCallback.kt
```diff
@@ -0,0 +1,43 @@
...
+import io.plaidapp.core.data.PlaidItem
+
+data class CapturedValues(
+    val result: List<PlaidItem>? = null,
+    val page: Int? = null,
+    val source: String? = null,
+    val error: String? = null
+)
+
+class AwesomeTestLoadSourceCallback : LoadSourceCallback {
```
ataulm: 🙈 My bad on the name.

How about splitting it into two? We never expect both callbacks to be called in the same execution flow.

It can also be moved to the `io.plaidapp.core.data` package (in the test sourceset), to match `LoadSourceCallback` since this isn't specific to `designernews`.

```kotlin
class SourceLoadedCallback : LoadSourceCallback {

    var params: Triple<List<PlaidItem>?, Int, String>? = null

    override fun sourceLoaded(result: List<PlaidItem>?, page: Int, source: String) {
        params = Triple(result, page, source)
    }

    override fun loadFailed(source: String) {
    }
}

class LoadFailedCallback : LoadSourceCallback {

    var source: String? = null

    override fun loadFailed(source: String) {
        this.source = source
    }

    override fun sourceLoaded(result: List<PlaidItem>?, page: Int, source: String) {
    }
}
```

core/src/test/java/io/plaidapp/core/designernews/domain/LoadStoriesUseCaseTest.kt
```diff
@@ -82,52 +78,33 @@ class LoadStoriesUseCaseTest {
...
-        }
+        // Given a callback where we check the validity of the data received
+        val callback = AwesomeTestLoadSourceCallback()
 
         // When loading stories
         loadStoriesUseCase(2, callback)
 
         // The correct callback was called
-        assertTrue(sourceLoadingFailed)
+        assertEquals(SourcesRepository.SOURCE_DESIGNER_NEWS_POPULAR, callback.failedValue())
```
ataulm:
```kotlin
// Then fails to load source
val sourceLoadingFailed = callback.failedValue() != null
assertTrue(sourceLoadingFailed)
```

arjunpola:
> I'll leave the rest for one of the project collaborators to review. Thanks again for picking up the issue ☺️

I have addressed all the reviews in this patch. Thank you so much @ataulm for all the feedback 😃

keyboardsurfer:
LGTM once @ataulm's remaining comments are addressed.

arjunpola:
> Please address & resolve remaining comments by @ataulm. LGTM otherwise.

My bad, did not realize I missed few.

PR #622, opened 2019-03-02
Fixes build - Updates repositories.gradle with latest dependencies needed to fix th…

florina-muntenescu:
…e build

PR #623, opened 2019-03-02
Makes the HomeActivity a lifecycle owner

florina-muntenescu:


florina-muntenescu:
This PR contains the fixes from #622 as well.

florina-muntenescu:
@ZacSweers any change you can take a look at this Dagger related PR? Thanks a lot in advance!

app/src/main/java/io/plaidapp/dagger/HomeModule.kt
```diff
@@ -40,20 +43,30 @@ import io.plaidapp.core.dribbble.data.api.model.Shot
...
 
-    @Provides
-    fun activity(): Activity = activity
+    @Binds
+    abstract fun homeActivityAsActivity(activity: HomeActivity): Activity
 
-    @Provides
-    fun columns(): Int = activity.resources.getInteger(R.integer.num_columns)
+    @Binds
+    abstract fun context(activity: Activity): Context
```
keyboardsurfer: Looks like we could remove this one. At least I don't see any references to raw `context`

florina-muntenescu: It's needed for `SharedPreferences`.

PR #624, opened 2019-03-03
Creates a connectivity checker lifecycle aware component

florina-muntenescu:


florina-muntenescu:
This should be merged after #623

core/src/main/java/io/plaidapp/core/ui/ConnectivityChecker.kt
```diff
@@ -0,0 +1,74 @@
...
+            monitoringConnectivity = false
+        }
+    }
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_RESUME)
+    fun startMonitoringConnectivity() {
+        val activeNetworkInfo = connectivityManager.activeNetworkInfo
+        val connected = activeNetworkInfo != null && activeNetworkInfo.isConnected
+        _connectedStatus.postValue(connected)
+        if (!connected) {
```
ataulm: If the device has no connectivity `onResume()`, but then obtains a connection - will observers of `connectedStatus` get the update?

app/src/main/java/io/plaidapp/dagger/HomeModule.kt
```diff
@@ -68,5 +70,13 @@ abstract class HomeModule(private val activity: Activity) {
         @JvmStatic
         @Provides
         fun isPocketInstalled(activity: Activity): Boolean = PocketUtils.isPocketInstalled(activity)
+
+        @JvmStatic
+        @Provides
+        fun connectivityChecker(activity: Activity): ConnectivityChecker {
+            val connectivityManager =
+                activity.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
```
nickbutcher: can we use the reified ext fun here instead of casting? https://developer.android.com/reference/kotlin/androidx/core/content/package-summary#(android.content.Context).getSystemService()

app/src/main/java/io/plaidapp/ui/HomeActivity.java
```diff
@@ -132,18 +128,31 @@
...
         inject(this, data -> {
             List<PlaidItem> items = PlaidItemsList.getPlaidItemsForDisplay(adapter.getItems(), data);
             adapter.setItems(items);
             checkEmptyState();
         });
 
+        getLifecycle().addObserver(connectivityChecker);
+        connectedStatus = connectivityChecker.getConnectedStatus();
+        connectedStatus.observe(this, connected -> {
+            if(connected) {
```
nickbutcher: formatting

app/src/main/java/io/plaidapp/ui/HomeActivity.java
```diff
@@ -118,11 +114,11 @@
     ImageButton fabPosting;
     GridLayoutManager layoutManager;
     private int columns;
-    boolean connected = true;
     private TextView noFiltersEmptyText;
-    private boolean monitoringConnectivity = false;
     private FilterAdapter filtersAdapter;
 
+    private LiveData<Boolean> connectedStatus;
```
nickbutcher: Should the checker hold this LD?

core/src/main/java/io/plaidapp/core/ui/ConnectivityChecker.kt
```diff
@@ -0,0 +1,79 @@
...
+            connectivityManager.unregisterNetworkCallback(connectivityCallback)
+            monitoringConnectivity = false
+        }
+    }
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_RESUME)
+    fun startMonitoringConnectivity() {
+        val connectivityManager =
+            appContext.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
+        val activeNetworkInfo = connectivityManager.activeNetworkInfo
```
pfmaggi: Nitpick: Maybe nice to set the correct type:
val activeNetworkInfo: NetworkInfo? = connectivityManager.activeNetworkInfo

given that getActoveNetworkInfo is not annotated as nullable but can return null.

core/src/main/java/io/plaidapp/core/ui/ConnectivityChecker.kt
```diff
@@ -0,0 +1,74 @@
...
+import androidx.lifecycle.OnLifecycleEvent
+
+/**
+ * Lifecycle aware connectivity checker that exposes the network connected status via a LiveData.
+ */
+class ConnectivityChecker(
+    private val connectivityManager: ConnectivityManager
+) : LifecycleObserver {
+
+    private var monitoringConnectivity = false
```
nickbutcher: Do we need this flag? We don't check it before registering the connectivity callback, only before unregistering.

core/src/main/java/io/plaidapp/core/ui/ConnectivityChecker.kt
pfmaggi: On a different note, why are you registering the callback only if there's no connectivity?
How are you going to be notified if you loose connectivity?

core/src/main/java/io/plaidapp/core/ui/ConnectivityChecker.kt
```diff
@@ -0,0 +1,79 @@
...
+    }
+
+    @OnLifecycleEvent(Lifecycle.Event.ON_RESUME)
+    fun startMonitoringConnectivity() {
+        val connectivityManager =
+            appContext.getSystemService(Context.CONNECTIVITY_SERVICE) as ConnectivityManager
+        val activeNetworkInfo = connectivityManager.activeNetworkInfo
+        connected = activeNetworkInfo != null && activeNetworkInfo.isConnected
+        _connectedStatus.postValue(connected)
+        if (!connected) {
```
pfmaggi: On a different note, why are you registering the callback only if there's no connectivity?
How are you going to be notified if you loose connectivity?

app/src/main/java/io/plaidapp/dagger/HomeModule.kt
ashdavies: It may not be safe to use these yet because `ContextCompat` returns incorrect values, last time I checked the fix hadn't yet made it into a stable release.

https://issuetracker.google.com/issues/116074337

nickbutcher: No need for compat as we're min SDK 23.

ashdavies: The extension function delegates to `ContextCompat`

https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/core/ktx/src/main/java/androidx/core/content/Context.kt#34

core/src/main/java/io/plaidapp/core/ui/ConnectivityChecker.kt
florina-muntenescu: Unfortunately we do. If we call `unregisterNetworkCallback` on a callback that's already been unregistered, `ConnectivityManager` throws an exception.

core/src/main/java/io/plaidapp/core/ui/ConnectivityChecker.kt
florina-muntenescu: now they will :)

core/src/main/java/io/plaidapp/core/ui/ConnectivityChecker.kt
florina-muntenescu: we won't handle that for now.

core/src/main/java/io/plaidapp/core/ui/ConnectivityChecker.kt
ataulm: ah I meant same as Pietro. is there harm in getting rid of the `if (!connected)`?

core/src/main/java/io/plaidapp/core/ui/ConnectivityChecker.kt
```diff
@@ -28,9 +29,17 @@ import androidx.lifecycle.OnLifecycleEvent
...
 /**
  * Lifecycle aware connectivity checker that exposes the network connected status via a LiveData.
+ *
+ * The loss of connectivity while the user scrolls through the feed should NOT be a blocker for the
+ * user.
+ *
+ * The loss of connectivity when the activity is resumed should be a blocker for the user
+ * (since we can't get feed items) - in onResume, we should get the connectivity status. If we
+ * are NOT connected then we register a listener and wait to be notified. Only once we are
+ * connected, we stop listening to connectivity.¬
```
ataulm: 👍

core/src/main/java/io/plaidapp/core/ui/ConnectivityChecker.kt
```diff
@@ -0,0 +1,87 @@
...
+ */
+class ConnectivityChecker(
+    private val connectivityManager: ConnectivityManager?
```
manuelvicnt: Should we allow `ConnectivityManager` to be null? IMO we shouldn't. The code in the class will always run in `onResume` / `onPause` and will keep posting the `No connectivity` value. The callback will never get executed plus we're wasting CPU and memory. WDYT?

app/src/main/java/io/plaidapp/ui/HomeActivity.java
```diff
@@ -132,18 +125,31 @@
...
         setContentView(R.layout.activity_home);
         bindResources();
+
         inject(this, data -> {
             List<PlaidItem> items = PlaidItemsList.getPlaidItemsForDisplay(adapter.getItems(), data);
             adapter.setItems(items);
             checkEmptyState();
         });
 
+        getLifecycle().addObserver(connectivityChecker);
```
manuelvicnt: add null check, please. `connectivityChecker` might be null if the service is not available

manuelvicnt: Java... 🙄

core/src/main/java/io/plaidapp/core/ui/ConnectivityChecker.kt
manuelvicnt: Taking a look at the `@Provider` we're not creating it if the `ConnectivityManager` is null so we can remove the question mark

app/src/main/java/io/plaidapp/ui/HomeActivity.java
```diff
@@ -576,7 +566,8 @@ void checkEmptyState() {
         if (adapter.getDataItemCount() == 0) {
             // if grid is empty check whether we're loading or if no filters are selected
             if (sourcesRepository.getActiveSourcesCount() > 0) {
-                if (connected) {
+                Boolean connected = connectivityChecker.getConnectedStatus().getValue();
```
manuelvicnt: Null check here as well. Is there any possibility that this code is executed before the `connectivityChecker` has posted a value?

florina-muntenescu: i don't think so, but since livedata values can be null, when converting this to kotlin, it will be nullable.

Issue #630, opened 2019-03-05
Add DataBinding to About 


florina-muntenescu:

manuelvicnt:
This was fixed in #633

PR #633, opened 2019-03-06
Add DataBinding to About screen

manuelvicnt:
Fixes issue #630 - Add DataBinding to About

Tested on an emulator and everything worked so far
All tests pass

### DIFF
--- a/about/src/main/res/layout/activity_about.xml
+++ b/about/src/main/res/layout/activity_about.xml
@@ -16,29 +16,32 @@
   ~
   -->
 
-<io.plaidapp.core.ui.widget.ElasticDragDismissFrameLayout
-    xmlns:android="http://schemas.android.com/apk/res/android"
-    xmlns:tools="http://schemas.android.com/tools"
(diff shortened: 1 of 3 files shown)


about/src/main/res/layout/activity_about.xml
```diff
@@ -16,29 +16,36 @@
...
-        android:layout_marginTop="@dimen/padding_normal"
-        app:pageIndicatorColor="@color/page_indicator_dark"
-        app:currentPageIndicatorColor="@color/page_indicator_dark_selected" />
-
-</io.plaidapp.core.ui.widget.ElasticDragDismissFrameLayout>
+<layout xmlns:android="http://schemas.android.com/apk/res/android"
+        xmlns:tools="http://schemas.android.com/tools"
+        xmlns:app="http://schemas.android.com/apk/res-auto">
+
+    <data>
```
florina-muntenescu: do we need the <data> tags?

manuelvicnt: We don't. Left it there in case we need it in the future but yeah, you're right. Removing it now!

Issue #636, opened 2019-03-06
Designer News Login fails


pfmaggi:
The current implementation in master fails to recognize the access_toker returned by Designer News.

Body always seen as zero size and login fails.

florina-muntenescu:
Looks like there are multiple ways to handle multi-converters. 
1. Create annotations and based on them, tell retrofit what to use (see how [here](https://speakerdeck.com/jakewharton/making-retrofit-work-for-you-ohio-devfest-2016))
2. Create another service for DesignerNewsSearch

I don't really like the annotations solution because we only need the converter for Search.
So, I'd suggest just breaking the search in another service, and only that one to use the `DesignerNewsSearchConverter`.

Happy to get other options on how to fix this as well.

ataulm:
d'oh I didn't read this. I went with option number one because we had [another search service before](https://github.com/nickbutcher/plaid/pull/604#discussion-diff-252928333R27) and we got rid of it.

I think option 1 is legit - we use it for the de-enveloping converter and it's a lot smaller than pulling another service out.

PR #638, opened 2019-03-06
Temporarly disable DN Search Converter to fix issue #636.

pfmaggi:
This is a quick fix to be able to be able to login into Designer News, while a better fix is implemented.


florina-muntenescu:
Issue fixed meanwhile by #642 
Thanks for investigating this, @pfmaggi!

PR #642, opened 2019-03-08
[636] Fix designer news converters

### LINKED ISSUE #636: Designer News Login fails

ataulm:
Fixes #636.

My bad, I didn't implement the search converter very well. Retrofit will go through the converters in order, looking for one that says it can convert the given type:

```
class Factory : Converter.Factory() {

    override fun responseBodyConverter(type: Type?, annotations: Array<Annotation>?, retrofit: Retrofit?): Converter<ResponseBody, *>? {
        return DesignerNewsSearchConverter
    }
}
```

It's meant to return `null` if it can't convert the response, but we return the converter without checking if it's suitable.

Since `DesignerNewsService` is handling both JSON responses (API) *and* HTML responses (website search results), we have to be more distinguishing.

There's two fixes:

- use a different service for the HTML responses `DesignerNewsSearchService`. This will only contain`DesignerNewsSearchConverter`, and not the `GsonConverterFactory`. It will only be used for searching.
- annotate the API class to help us determine the response type implement `responseBodyConverter` properly!

I went with the second option because it was quicker and I thought we were trying to avoid a separate service just for search 🙈 

fixed


core/src/main/java/io/plaidapp/core/designernews/data/api/HtmlResponse.kt
```diff
@@ -16,26 +16,9 @@
...
-import kotlinx.coroutines.Deferred
-import retrofit2.Response
-import retrofit2.http.GET
-import retrofit2.http.Query
-
 /**
- * Fake-API for searching Designer News
+ * Used to differentiate Designer News search response (HTML) from the JSON API responses.
  */
-interface DesignerNewsSearchService {
```
ataulm: I should have removed this as part of [this commit](https://github.com/nickbutcher/plaid/commit/2c0119741c3f0c144c23e0ef9646d9e3fbcb752d) but I didn't.

It was unused, except for the `ENDPOINT` in the companion object.

Git is confusedly marking this as a rename - `HtmlResponse` is a completely separate file I added.

core/src/main/java/io/plaidapp/core/designernews/data/api/DesignerNewsSearchConverter.kt
```diff
@@ -42,12 +42,26 @@ object DesignerNewsSearchConverter : Converter<ResponseBody, List<String>> {
...
+                    }
+                }
+            }
+            return false
         }
     }
 
     override fun convert(body: ResponseBody): List<String> {
-        val searchResults = Jsoup.parse(body.string(), DesignerNewsSearchService.ENDPOINT)
+        val searchResults = Jsoup.parse(body.string(), DesignerNewsService.ENDPOINT)
```
ataulm: These have the same value. The former has been removed from the codebase in this PR, because the rest of the class was unused.

ataulm:
I learned about this from this article, it was really simple and clear - https://medium.com/@oldergod/how-i-got-the-best-of-retrofit-2297cb56a40f

florina-muntenescu:
The `@Html` annotation would suggest that whenever we have another endpoint that returns Html, then we can use the `DesignerNewsSearchConverter`, which is not true, since the converter is specific only to this endpoint. 
As stated in #636 I think another Service might be a better solution to this. I know I was the one against it in the first implementation of the Designer News search functionality, but, the converter is a good enough argument for me to support the separation.
Opened to other suggestions as well!

ataulm:
>I know I was the one against it in the first implementation of the Designer News search functionality

Neither of us realised the gain before so this shouldn't weigh in 🙏

I'm not married to this but I think it's more practical for now:

- it's smaller
- it's done 😬
- the difference between API responses is explicitly marked so it's clear that `search` is distinct from other functions in DNService

If the only beef is the name, we can rename it to `@DesignerNewsSearchResponse` or anything.

I think the reason you wrote for not having the annotation in the issue could also apply to not having a separate service ("don't really like the annotations solution because we only need the converter for Search.")

I mean, if you want to go for a separate service, half the code is already there because I forgot to delete it in the last PR 😂

florina-muntenescu:
Ok. Let's go for `@DesignerNewsSearch` annotation then. 
Could you create another PR cleaning up the `DesignerNewsSearchService`?

ataulm:
This PR deleted `DesignerNewsSearchService` already.

core/src/main/java/io/plaidapp/core/designernews/data/api/DesignerNewsSearchConverter.kt
```diff
@@ -42,12 +42,26 @@ object DesignerNewsSearchConverter : Converter<ResponseBody, List<String>> {
...
             retrofit: Retrofit?
         ): Converter<ResponseBody, *>? {
-            return DesignerNewsSearchConverter
+            if (annotations.anyMatch { it is DesignerNewsSearch }) {
+                return DesignerNewsSearchConverter
+            }
+            return null
+        }
+
+        private fun Array<Annotation>?.anyMatch(condition: (Annotation) -> Boolean): Boolean {
```
nickbutcher: Could we use `any` here instead?

http://kotlinlang.org/api/latest/jvm/stdlib/kotlin.collections/any.html

ataulm: yess 👌 

commit d71ffbc45ff5db4ee5ceeafa6e49f161dc0f0d6b (HEAD -> fix-designer-news-converters, ataulm/fix-designer-news-converters)
Author: Ataul Munim <hello@ataulm.com>
Date:   Sat Mar 9 09:34:58 2019 +0000

    Replace custom any with kotlin's built-in one

ataulm:
I can't reproduce it locally, running the same (Android build) commands as the CI workflow:

`./gradlew check assemble assembleAndroidTest`

I've seen AAPT errors before being OS specific. Maybe we can update the Android Gradle Plugin from `3.4.0-beta04` to `3.4.0-rc01` or `3.5.*`.

nickbutcher:
Can you try triggering a rerun?

ataulm:
I don't have permissions to trigger a workflow except via PRs.

This one passes with an updated AGP - https://github.com/nickbutcher/plaid/pull/645

PR #643, opened 2019-03-09
[636] Fix error when logging into DN service

ataulm:
Alternate approach to #642 (don't merge both 😬)

- the login API response was being converted by the DN Search Converter instead of the GSON converter
- split them into two services, each one only knows about the correct converter


ataulm:
going with #642

PR #644, opened 2019-03-09
Cleaning FeedAdapter - step 2

florina-muntenescu:
Removing DataLoadingSubject from FeedAdapter and moving the logic to the Home and Search ViewModels

### DIFF
--- a/app/src/main/java/io/plaidapp/ui/HomeActivity.java
+++ b/app/src/main/java/io/plaidapp/ui/HomeActivity.java
@@ -72,7 +72,7 @@ import io.plaidapp.core.designernews.data.poststory.PostStoryService;
 import io.plaidapp.core.designernews.data.stories.model.Story;
 import io.plaidapp.core.dribbble.data.api.model.Shot;
 import io.plaidapp.core.ui.ConnectivityChecker;
-import io.plaidapp.core.ui.FeedAdapter;
+import io.plaidapp.core.feed.FeedAdapter;
 import io.plaidapp.core.ui.HomeGridItemAnimator;
 import io.plaidapp.core.ui.PlaidItemsList;
 import io.plaidapp.core.ui.filter.FilterAdapter;
@@ -142,7 +142,7 @@ public class HomeActivity extends FragmentActivity {
 
         boolean pocketInstalled = PocketUtils.isPocketInstalled(this);
 
-        adapter = new FeedAdapter(this, dataManager, columns, pocketInstalled);
+        adapter = new FeedAdapter(this, columns, pocketInstalled);
 
         if(connectivityChecker != null) {
             getLifecycle().addObserver(connectivityChecker);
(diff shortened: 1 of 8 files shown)


app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -303,12 +308,40 @@ class HomeViewModelTest {
         assertEquals(inactiveSource.key, source)
     }
 
+    @Test
+    fun dataLoading() {
+        // Given a view model
+        val homeViewModel = createViewModel()
+        Mockito.verify(dataManager).registerCallback(capture(dataLoadingCallback))
```
nickbutcher: Static import?

core/src/main/java/io/plaidapp/core/feed/FeedUiModel.kt
```diff
@@ -0,0 +1,30 @@
...
+import io.plaidapp.core.data.PlaidItem
+
+/**
+ * UI model for feed data
+ */
+data class FeedUiModel(
+    val items: List<PlaidItem>
+)
+
+data class FeedProgressUiModel(
```
nickbutcher: Should we consider enabling experimental inline classes?

search/src/main/java/io/plaidapp/search/ui/SearchViewModel.kt
```diff
@@ -31,19 +33,34 @@ import io.plaidapp.search.domain.SearchDataManager
...
+
     private val onDataLoadedCallback: OnDataLoadedCallback<List<PlaidItem>> =
         object : OnDataLoadedCallback<List<PlaidItem>> {
             override fun onDataLoaded(data: List<PlaidItem>) {
-                _searchResults.value = Event(data)
+                _searchResults.value = Event(FeedUiModel(data))
             }
         }
 
+    private val dataLoadingCallbacks = object : DataLoadingSubject.DataLoadingCallbacks {
```
nickbutcher: Worth factoring this out somewhere?

core/src/main/java/io/plaidapp/core/feed/FeedUiModel.kt
florina-muntenescu: as soon as this is not used in java classes, yes.

search/src/main/java/io/plaidapp/search/ui/SearchViewModel.kt
florina-muntenescu: In this PR I wanted to move things out of the `FeedAdapter`. We shouldn't have such a callback at all in our app so, I expect that we will be able to remove it completely in a future PR.

florina-muntenescu:
@nickbutcher - ready for another 👀

app/src/main/java/io/plaidapp/ui/HomeActivity.java
```diff
@@ -174,6 +174,14 @@ protected void onCreate(Bundle savedInstanceState) {
                     checkEmptyState();
                 });
 
+        viewModel.getFeedProgress().observe(this, feedProgressUiModel -> {
```
nickbutcher: Move trailing lambda out of parens?

florina-muntenescu: Not in Java :)

PR #645, opened 2019-03-09
[test] Update AGP

ataulm:
Cryptic error when raising #642 :

* What went wrong:
Execution failed for task ':bypass:verifyReleaseResources'.
> 1 exception was raised by workers:
  com.android.builder.internal.aapt.v2.Aapt2InternalException: AAPT2 aapt2-3.4.0-beta04-5252756-linux Daemon #0: Unexpected error during compile '/home/circleci/plaid/third_party/bypass/build/intermediates/res/merged/release/values/values.xml', attempting to stop daemon.
  This should not happen under normal circumstances, please file an issue if it does.

Testing updating the AGP version.


ataulm:
Seems to work with updated AGP.

PR #646, opened 2019-03-10
Update AGP to 3.4.0-rc01

ataulm:
Updates to the latest RC. It seems to resolve an issue in AAPT which causes #642 to fail.

There's `3.5.0-alpha06` but it'd also require a Gradle update (happy to do this, if preferred).


ataulm:
#642 is already green 🤔

PR #648, opened 2019-03-10
Separate the response object from product hunt from the domain object

florina-muntenescu:


core/src/main/java/io/plaidapp/core/dagger/DataManagerModule.kt
```diff
@@ -42,13 +42,13 @@ class DataManagerModule {
         onDataLoadedCallback: OnDataLoadedCallback<List<PlaidItem>>,
         loadStoriesUseCase: LoadStoriesUseCase,
         searchStoriesUseCase: SearchStoriesUseCase,
-        productHuntRepository: ProductHuntRepository,
+        loadPostsUseCase: LoadPostsUseCase,
```
nickbutcher: Aren't we dropping the `UseCase` suffix on param names?

core/src/main/java/io/plaidapp/core/data/DataManager.java
```diff
@@ -203,7 +204,7 @@ private void loadDribbbleSearch(final Source.DribbbleSearchSource source, final
 
     private void loadProductHunt(final int page) {
         // this API's paging is 0 based but this class (& sorting) is 1 based so adjust locally
-        productHuntRepository.loadProductHuntData(
+        loadPosts.invoke(
```
nickbutcher:
```suggestion
        loadPosts(
```

core/src/main/java/io/plaidapp/core/producthunt/data/api/ProductHuntRepository.kt
```diff
@@ -34,38 +29,14 @@ class ProductHuntRepository(
...
-        onSuccess: (List<Post>) -> Unit,
-        onError: (String) -> Unit
-    ) = scope.launch(dispatcherProvider.io) {
-        val result = remoteDataSource.loadData(page)
-        withContext(dispatcherProvider.main) {
-            when (result) {
-                is Result.Success -> onSuccess(result.data)
-                is Result.Error -> onError(result.exception.toString())
-            }.exhaustive
+    suspend fun loadPosts(page: Int): Result<GetPostsResponse> =
```
nickbutcher: Lets not use an expression for multi-line method.

core/src/main/java/io/plaidapp/core/producthunt/data/api/ProductHuntService.kt
```diff
@@ -28,9 +27,8 @@ import retrofit2.http.Query
  */
 interface ProductHuntService {
 
-    @EnvelopePayload("posts")
     @GET("v1/posts")
-    fun getPosts(@Query("days_ago") page: Int): Deferred<Response<List<Post>>>
+    fun getPostsAsync(@Query("days_ago") page: Int): Deferred<Response<GetPostsResponse>>
```
nickbutcher: Is it worth the `Async` suffix if we're not differentiating from a sync method? Isn't the type sufficient here? Is this consistent with elsewhere in the project?

core/src/main/java/io/plaidapp/core/producthunt/domain/LoadPostsUseCase.kt
```diff
@@ -0,0 +1,79 @@
...
+import javax.inject.Inject
+
+/**
+ * Class that knows how to load posts
+ */
+class LoadPostsUseCase @Inject constructor(
+    private val productHuntRepository: ProductHuntRepository,
+    private val dispatcherProvider: CoroutinesDispatcherProvider
+) {
+    private var parentJob = Job()
```
nickbutcher: Shouldn't we use a `SupervisorJob` here so that individual coroutines can fail without failing the whole thing?

hs_err_pid7739.log
```diff
@@ -0,0 +1,620 @@
+#
```
nickbutcher: Should this be checked in???

core/src/main/java/io/plaidapp/core/data/DataManager.java
florina-muntenescu: ummm... in Java?

core/src/main/java/io/plaidapp/core/producthunt/data/api/ProductHuntService.kt
florina-muntenescu: Without `Async` I get the following warning:

> Function returning Deferred with a name that does not end with Async less... (⌘F1) 
Inspection info: This inspection reports functions with kotlinx.coroutines.Deferred result. Functions which use Deferred as return type should have a name with suffix Async. Otherwise, it's recommended to turn a function into a suspend function and unwrap Deferred.

We can also live with this warning until we make all Service functions suspending, when we won't get this warning anyway. What do you think?

hs_err_pid7739.log
florina-muntenescu: wooooops

PR #649, opened 2019-03-11
Refactor Dagger components to use CoreComponent

manuelvicnt:
Tested in an emulator. Would be nice to test in a real device and try some features like login.

## **State of the project before the refactoring**
A Component can be useful for:
- Be injected to an app component.
- Inject modules so that other components that are Subcomponents of it have those modules available.
- Expose dependencies so that other Components that depends on this one can access them.

Before the refactoring, you could find modules in a component defined multiple times. Why?

Since Plaid doesn't structure the Dagger graph with SubComponents. CoreComponent wasn't used as expected. It was never injected to an app component, imported 6 modules and didn't expose any dependency. Those 6 modules created Providers in the CoreComponentDagger generated file that weren't used by any other Component that had CoreComponent as a dependency because those Providers weren't exposed. How can you expose an Object from a Component that can be consumed by others? Exposing the Object in the interface.  

```
interface CoreComponent {
    ...
    fun provideGson(): Gson
}
```

With that, a Component that adds CoreComponent as a dependency can use Gson without having to add it in its list of modules.

Since the XComponent didn't take advantage of the dependencies provided by CoreComponent, Modules were a bit tangled-up because they had to include other Modules (even if they were available elsewhere) just to satisfy its own dependencies.

**Also**, if a Component includes XModule and YModule, if YModule needs XModule, it's not necessary for YModule to include XModule since the Component already includes it. There were few cases like this in the project that have been refactored now. 

## **What is this PR about?**
Just to keep the PR small and simple, it just cleans up the Modules with redundant includes and the Components to use the CoreComponent properly as a dependency, not as a SubComponent. The CoreComponent has been simplified to contain just the Modules that are needed by the rest of the Components.  

## **Next steps**
- Make objects live as long as the component they're in by using Scopes. Issue #573 
- Refactor some Modules that depend on a lot of modules (e.g. SearchModule). 

## **Component lists with Modules**

**CoreComponent**
- CoreDataModule::class

**AboutComponent (Doesn’t add CoreComponent)**
- MarkDownModule
- AboutActivityModule

**DribbbleComponent (adds Core)**

**HomeComponent (adds Core)**
- HomeModule
- - DataManagerModule::class,
- - - DesignerNewsDataModule::class
- - - ProductHuntModule::class
- - SourcesRepositoryModule::class
- - DribbbleDataModule::class
- - OnDataLoadedModule::class
- SharedPreferencesModule

**UpvoteStoryServiceComponent (adds Core)**
- UpvoteStoryServiceModule::class,
- - DesignerNewsDataModule::class
- DesignerNewsDataModule
- SharedPreferencesModule

**StoryComponent (adds Core)**
- StoryModule
- DesignerNewsDataModule::class,
- DataModule::class,
- MarkdownModule::class
- SharedPreferencesModule::class

**LoginComponent (adds Core)**
- DesignerNewsDataModule::class,
- SharedPreferencesModule::class

**SearchComponent (adds Core)**
- SearchModule
- - DribbbleDataModule::class,
- - DesignerNewsDataModule::class,
- - SharedPreferencesModule::class

### DIFF
--- a/app/src/main/java/io/plaidapp/ui/PlaidApplication.kt
+++ b/app/src/main/java/io/plaidapp/ui/PlaidApplication.kt
@@ -29,8 +29,7 @@ class PlaidApplication : Application() {
 
     private val coreComponent: CoreComponent by lazy {
         DaggerCoreComponent
-            .builder()
-            .build()
+            .create()
     }
 
     companion object {
--- a/app/src/main/java/io/plaidapp/ui/PlaidApplication.kt
+++ b/app/src/main/java/io/plaidapp/ui/PlaidApplication.kt
@@ -21,7 +21,6 @@ import android.app.Application
 import android.content.Context
 import io.plaidapp.core.dagger.CoreComponent
 import io.plaidapp.core.dagger.DaggerCoreComponent
-import io.plaidapp.core.dagger.MarkdownModule
 
 /**
  * Io and Behold
@@ -31,7 +30,6 @@ class PlaidApplication : Application() {
     private val coreComponent: CoreComponent by lazy {
         DaggerCoreComponent
             .builder()
-            .markdownModule(MarkdownModule(resources.displayMetrics))
             .build()
     }
 
--- a/core/src/main/java/io/plaidapp/core/dagger/CoreDataModule.kt
+++ b/core/src/main/java/io/plaidapp/core/dagger/CoreDataModule.kt
@@ -24,6 +24,7 @@ import io.plaidapp.core.BuildConfig
 import okhttp3.OkHttpClient
 import okhttp3.logging.HttpLoggingInterceptor
 import retrofit2.converter.gson.GsonConverterFactory
+import javax.inject.Named
 
 /**
  * Dagger module to provide core data functionality.
@@ -32,6 +33,7 @@ import retrofit2.converter.gson.GsonConverterFactory
 class CoreDataModule {
 
     @Provides
+    @Named("coreOkHttpClient")
     fun provideOkHttpClient(interceptor: HttpLoggingInterceptor): OkHttpClient =
         OkHttpClient.Builder().addInterceptor(interceptor).build()
 
--- a/core/src/main/java/io/plaidapp/core/dagger/CoreDataModule.kt
+++ b/core/src/main/java/io/plaidapp/core/dagger/CoreDataModule.kt
@@ -24,7 +24,6 @@ import io.plaidapp.core.BuildConfig
 import okhttp3.OkHttpClient
 import okhttp3.logging.HttpLoggingInterceptor
 import retrofit2.converter.gson.GsonConverterFactory
-import javax.inject.Named
 
 /**
  * Dagger module to provide core data functionality.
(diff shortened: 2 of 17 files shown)


manuelvicnt:
cc @ZacSweers in case you want to take a look

ZacSweers:
See my comment on https://github.com/nickbutcher/plaid/pull/650#issuecomment-471946622 for qualifiers :). 

In general this is a positive step! I do think subcomponents would be a good use here eventually though, and personally think removing the LocalApi pattern is exposing those providers as a result when the original intent was to use a private annotation that makes it impossible access them from outside that annotation's visibility. I'd also advise against qualifying root types unnecessarily, as that's sort of a default and kind of noisy IMO. Qualifiers should be to differentiate from an existing standard or primitive type, not just to denote if something is coming from a specific scope. That's what scope annotations are for! 🙂

app/src/main/java/io/plaidapp/ui/PlaidApplication.kt
```diff
@@ -31,7 +30,6 @@ class PlaidApplication : Application() {
     private val coreComponent: CoreComponent by lazy {
         DaggerCoreComponent
```
ronshapiro: Use `DaggerCoreComponent.create()` if there's no modules to be passed

core/src/main/java/io/plaidapp/core/dagger/CoreDataModule.kt
```diff
@@ -32,6 +33,7 @@ import retrofit2.converter.gson.GsonConverterFactory
 class CoreDataModule {
 
     @Provides
+    @Named("coreOkHttpClient")
```
ronshapiro: Let's define a custom qualifier here. `@Named` is brittle if you misspell the string.

core/src/main/java/io/plaidapp/core/dagger/designernews/Injector.kt
```diff
@@ -26,9 +28,16 @@ import io.plaidapp.core.designernews.data.votes.UpvoteStoryService
  * Injector for [UpvoteStoryService].
  */
 
+private val coreComponent: CoreComponent by lazy {
+    DaggerCoreComponent
```
ronshapiro: Shouldnt this be using the same instance from PlaidApplication? Otherwise they wont' actually share any instances

designernews/src/main/java/io/plaidapp/designernews/dagger/StoryComponent.kt
```diff
@@ -21,12 +21,21 @@ import io.plaidapp.core.dagger.BaseActivityComponent
...
+@Component(
+    modules = [
+        SharedPreferencesModule::class,
+        MarkdownModule::class,
+        DesignerNewsDataModule::class,
+        DataModule::class,
+        StoryModule::class],
+    dependencies = [CoreComponent::class]
+)
 interface StoryComponent : BaseActivityComponent<StoryActivity> {
```
ronshapiro: Should this be a subcomponent of something in the designernews module?

core/src/main/java/io/plaidapp/core/dagger/designernews/Injector.kt
manuelvicnt: It should. However, this is in the core module and it's not part of the app module so it cannot access the CoreComponent. We'll try to refactor this later

designernews/src/main/java/io/plaidapp/designernews/dagger/StoryComponent.kt
manuelvicnt: Absolutely. This needs a clean up that will be coming in upcoming PRs

Issue #537, opened 2018-10-26
[Dagger] Use @IsPocketInstalled dagger qualifier


florina-muntenescu:
From [PR comment](https://github.com/nickbutcher/plaid/pull/534#discussion_r228313545).

PR #650, opened 2019-03-12
[Dagger] Use @IsPocketInstalled dagger qualifier

pt2121:
re #537 

* adding `IsPocketInstalled` qualifier
* ~adding `FeedColumns` qualifier~
* ~injecting `FeedAdapter`~

### DIFF
--- a/app/src/main/java/io/plaidapp/dagger/HomeModule.kt
+++ b/app/src/main/java/io/plaidapp/dagger/HomeModule.kt
@@ -29,6 +29,7 @@ import io.plaidapp.R
 import io.plaidapp.core.dagger.DataManagerModule
 import io.plaidapp.core.dagger.SourcesRepositoryModule
 import io.plaidapp.core.dagger.dribbble.DribbbleDataModule
+import io.plaidapp.core.dagger.qualifier.IsPocketInstalled
 import io.plaidapp.core.data.pocket.PocketUtils
 import io.plaidapp.core.ui.ConnectivityChecker
 import io.plaidapp.ui.HomeActivity
@@ -63,6 +64,7 @@ abstract class HomeModule {
         @Provides
         fun columns(activity: Activity): Int = activity.resources.getInteger(R.integer.num_columns)
 
+        @IsPocketInstalled
         @JvmStatic
         @Provides
         fun isPocketInstalled(activity: Activity): Boolean = PocketUtils.isPocketInstalled(activity)
--- a/app/src/main/java/io/plaidapp/ui/HomeActivity.java
+++ b/app/src/main/java/io/plaidapp/ui/HomeActivity.java
@@ -63,8 +63,8 @@ import androidx.recyclerview.widget.RecyclerView;
 import com.bumptech.glide.integration.recyclerview.RecyclerViewPreloader;
 import com.bumptech.glide.util.ViewPreloadSizeProvider;
 import io.plaidapp.R;
+import io.plaidapp.core.dagger.qualifier.IsPocketInstalled;
 import io.plaidapp.core.data.PlaidItem;
-import io.plaidapp.core.data.pocket.PocketUtils;
 import io.plaidapp.core.data.prefs.SourcesRepository;
 import io.plaidapp.core.designernews.data.poststory.PostStoryService;
 import io.plaidapp.core.designernews.data.stories.model.Story;
@@ -124,6 +124,10 @@ public class HomeActivity extends FragmentActivity {
     @Inject
     HomeViewModel viewModel;
 
+    @IsPocketInstalled
+    @Inject
+    boolean pocketInstalled;
+
     @Override
     protected void onCreate(Bundle savedInstanceState) {
         super.onCreate(savedInstanceState);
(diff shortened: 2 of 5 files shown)


app/src/main/java/io/plaidapp/ui/HomeActivity.java
```diff
@@ -113,7 +113,9 @@
     private int columns;
     private TextView noFiltersEmptyText;
     private FilterAdapter filtersAdapter;
-    private FeedAdapter adapter;
+
+    @Inject
```
florina-muntenescu: We don't need to inject the FeedAdapter, it's ok to let the HomeActivity create it.

core/src/main/java/io/plaidapp/core/dagger/qualifier/FeedColumns.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package io.plaidapp.core.dagger.qualifier
+
+import javax.inject.Qualifier
+import kotlin.annotation.AnnotationRetention.BINARY
+
+@Retention(BINARY)
+@Qualifier
+annotation class FeedColumns
```
florina-muntenescu: new line

core/src/main/java/io/plaidapp/core/dagger/qualifier/IsPocketInstalled.kt
```diff
@@ -0,0 +1,24 @@
...
+ */
+
+package io.plaidapp.core.dagger.qualifier
+
+import javax.inject.Qualifier
+import kotlin.annotation.AnnotationRetention.BINARY
+
+@Retention(BINARY)
+@Qualifier
+annotation class IsPocketInstalled
```
florina-muntenescu: new line

core/src/main/java/io/plaidapp/core/feed/FeedAdapter.java
```diff
@@ -101,9 +104,10 @@
     private List<PlaidItem> items;
     private boolean showLoadingMore = false;
 
+    @Inject
```
florina-muntenescu: don't inject the FeedAdapter.

search/src/main/java/io/plaidapp/search/ui/SearchActivity.kt
```diff
@@ -90,7 +89,8 @@ class SearchActivity : AppCompatActivity() {
     private val transitions = SparseArray<Transition>()
     private var focusQuery = true
 
-    private lateinit var feedAdapter: FeedAdapter
+    @Inject
```
florina-muntenescu: As in home, don't inject the FeedAdapter, let the SearchActivity create it.

manuelvicnt:
I personally prefer the `@Named` annotation to solve this kind of problem. The naming can be more concise and less code to write. I'm using it in #649 in case you want to take a look.

Should we move away from Qualifiers and use `@Named`?

florina-muntenescu:
> I personally prefer the `@Named` annotation to solve this kind of problem. The naming can be more concise and less code to write. I'm using it in #649 in case you want to take a look.
> 
> Should we move away from Qualifiers and use `@Named`?

@pt2121 - let's use this approach. Thanks!

ZacSweers:
Commented on the linked PR as well, but Named should be the exception and not the standard IMO.

Custom qualifiers can:
- be documented
- contain extra fields
- be stripped by proguard via non-runtime retention
- don't require sharing and reusing a string constant for consistency (which Named conventionally should use to avoid errors)
- can have custom visibility (package/protected/private) to limit access

And personal preference - I find `@ForPocket isPocketInstalled: Boolean` nicer/simpler to read than `@Named("ForPocket") isPocketInstalled: Boolean` or `@Named(FOR_POCKET) isPocketInstalled: Boolean`. The qualifier is just one element - the name. With Named, it's an element and its value that you have to from at a glance.

app/src/main/java/io/plaidapp/dagger/HomeModule.kt
```diff
@@ -61,10 +63,12 @@ abstract class HomeModule {
     @Module
     companion object {
 
+        @FeedColumns
```
florina-muntenescu: Let's stick with the initial scope of #537 and just use the qualifier for `IsPocketInstalled`. The columns won't be retrieved anywhere else than in the activity and we're not injecting them in the adapter, so I don't think there's a need to inject it.

manuelvicnt:
@ZacSweers made really good points and also, @ronshapiro in the other PR (#649). Seems we should go for that approach. So it's ok if you leave the implementation as it is :) I'll change mine! Thank you

pt2121:
thanks y'all for the feedback. updated my PR based on the comments

pt2121:
@florina-muntenescu , could you take another look when having a chance?
I think I addressed all comments. thx!

florina-muntenescu:
Looks good to me. 
@ZacSweers would you mind taking a look as well, please?

search/src/main/java/io/plaidapp/search/ui/SearchActivity.kt
```diff
@@ -95,6 +95,10 @@ class SearchActivity : AppCompatActivity() {
     @Inject
     lateinit var viewModel: SearchViewModel
 
+    @JvmField
+    @field:[Inject IsPocketInstalled]
```
ZacSweers: I always forget you can do this notation, neat!

florina-muntenescu:
Thanks for the PR @pt2121 !

PR #651, opened 2019-03-12
Clean up the feed in HomeActivity - Step 3

florina-muntenescu:
Move logic from HomeActivity to ViewModel:
* Handling of the removal/deactivation of a source
* Requesting data loading at every activity onCreate - moved to init of the ViewModel
* Move check for blank query to HomeViewModel

Update the weighting algorithm to allow multiple types in the list.


app/src/main/java/io/plaidapp/ui/HomeViewModel.kt
```diff
@@ -88,10 +99,15 @@ class HomeViewModel(
         }
     }
 
+    // TODO - find a better solution
```
nickbutcher: Issue number

app/src/main/java/io/plaidapp/ui/HomeViewModel.kt
```diff
@@ -88,10 +99,15 @@ class HomeViewModel(
...
+    var columns = 2
+
     init {
         sourcesRepository.registerFilterChangedCallback(filtersChangedCallbacks)
+        dataManager.setOnDataLoadedCallback(onDataLoadedCallback)
```
nickbutcher: So we have `onDataLoadedCallbacks` and `dataLoadingCallbacks` objects in this class. Wonder if we could clarify the naming to better differentiate them?

app/src/test/java/io/plaidapp/TestData.kt
```diff
@@ -0,0 +1,92 @@
...
+import io.plaidapp.core.dribbble.data.api.model.Images
+import io.plaidapp.core.dribbble.data.api.model.Shot
+import io.plaidapp.core.dribbble.data.api.model.User
+import io.plaidapp.core.producthunt.data.api.model.Post
+import io.plaidapp.core.ui.filter.SourceUiModel
+import java.util.Date
+import java.util.GregorianCalendar
+
+val designerNewsSource = Source.DesignerNewsSearchSource(
+    "query",
```
nickbutcher: Named params for readability.

nickbutcher: Throughout for test data.

app/src/test/java/io/plaidapp/TestData.kt
```diff
@@ -0,0 +1,92 @@
...
+    avatarUrl = "www.prettyplaid.nb"
+)
+
+val shot = Shot(
+    id = 1L,
+    title = "Foo Nick",
+    description = "",
+    images = Images(),
+    user = player
+).also {
```
nickbutcher:
```suggestion
).apply {
```

app/src/test/java/io/plaidapp/TestData.kt
```diff
@@ -0,0 +1,92 @@
...
+    downvotes = emptyList()
+)
+
+val story = Story(
+    id = storyId,
+    title = "Plaid 2.0 was released",
+    createdAt = createdDate,
+    userId = userId,
+    links = storyLinks
+).also {
```
nickbutcher:
```suggestion
).apply {
```

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -126,7 +134,11 @@ class HomeViewModelTest {
         val homeViewModel = createViewModel()
 
         // When adding a Designer News source
-        homeViewModel.addSources(designerNewsSource.query, isDribbble = false, isDesignerNews = true)
+        homeViewModel.addSources(
+            designerNewsSource.query,
```
nickbutcher:
```suggestion
            query = designerNewsSource.query,
```

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -259,41 +271,42 @@ class HomeViewModelTest {
...
-        val homeViewModel = createViewModel()
+        // Given a view model with feed data
+        val homeViewModel = createViewModelWithFeedData(listOf(post, shot, story))
         verify(sourcesRepository).registerFilterChangedCallback(
             capture(filtersChangedCallback)
         )
+        val initialFeed = LiveDataTestUtil.getValue(homeViewModel.feed)
 
         // When an active source was changed
         val activeSource = Source.DribbbleSearchSource("dribbble", true)
```
nickbutcher: Can we copy `dribbbleSource` test data to make the pertinent data obvious? i.e.

`val activeSource = dribbbleSource.copy(active = true)`

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -302,9 +315,9 @@ class HomeViewModelTest {
         val inactiveSource = Source.DribbbleSearchSource("dribbble", false)
```
nickbutcher: Ditto above.

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -335,6 +348,39 @@ class HomeViewModelTest {
...
+    private fun createViewModelWithFeedData(feedData: List<PlaidItem>): HomeViewModel {
+        val homeViewModel = createViewModel()
+        verify(dataManager).setOnDataLoadedCallback(capture(dataLoadedCallback))
+
+        // When data loaded return feedData
+        dataLoadedCallback.value.onDataLoaded(feedData)
+
+        return homeViewModel
+    }
+
```
nickbutcher: We're not testing that popular items are expanded which happens when the feed changes. File an issue to come back to this?

core/src/main/java/io/plaidapp/core/data/DataManager.java
```diff
@@ -72,13 +71,15 @@ public DataManager(OnDataLoadedCallback<List<? extends PlaidItem>> onDataLoadedC
...
 
-    private void setOnDataLoadedCallback(
+    public void setOnDataLoadedCallback(
             OnDataLoadedCallback<List<? extends PlaidItem>> onDataLoadedCallback) {
         this.onDataLoadedCallback = onDataLoadedCallback;
     }
 
     private void onDataLoaded(List<? extends PlaidItem> data) {
-        onDataLoadedCallback.onDataLoaded(data);
+        if(onDataLoadedCallback != null) {
```
nickbutcher:
```suggestion
        if (onDataLoadedCallback != null) {
```

app/src/main/java/io/plaidapp/ui/HomeViewModel.kt
florina-muntenescu: I'm planning to remove at least one of them in a next PR. 
I don't think we should use callbacks at all. 
Is it worth changing it now?

app/src/test/java/io/plaidapp/TestData.kt
florina-muntenescu: added to the data classes.
Some others, like the Source here, is still java

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
florina-muntenescu: java :)

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
florina-muntenescu: #653 created

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
florina-muntenescu: would love to. but sources are still java

nickbutcher:
Thanks for the changes 👍

PR #656, opened 2019-03-13
Pass columns when getting the feed in Home

### LINKED ISSUE #654: Pass the columns to the VM when starting to observe the feed items

florina-muntenescu:
Fixes #654

### DIFF
--- a/app/src/main/java/io/plaidapp/ui/HomeActivity.java
+++ b/app/src/main/java/io/plaidapp/ui/HomeActivity.java
@@ -33,7 +33,6 @@ import android.text.Spannable;
 import android.text.SpannableStringBuilder;
 import android.text.Spanned;
 import android.text.SpannedString;
-import android.text.TextUtils;
 import android.text.style.ForegroundColorSpan;
 import android.text.style.ImageSpan;
 import android.transition.TransitionManager;
@@ -132,8 +131,6 @@ public class HomeActivity extends FragmentActivity {
 
         inject(this);
 
-        viewModel.setColumns(columns);
-
         boolean pocketInstalled = PocketUtils.isPocketInstalled(this);
 
         adapter = new FeedAdapter(this, columns, pocketInstalled);
@@ -172,7 +169,7 @@ public class HomeActivity extends FragmentActivity {
             }
         });
 
-        viewModel.getFeed().observe(this, feedUiModel -> {
+        viewModel.getFeed(columns).observe(this, feedUiModel -> {
             adapter.setItems(feedUiModel.getItems());
             checkEmptyState();
         });
@@ -492,7 +489,7 @@ public class HomeActivity extends FragmentActivity {
                     // actually add the story to the grid
                     Story newStory = intent.getParcelableExtra(PostStoryService.EXTRA_NEW_STORY);
 
-                    List<PlaidItem> items = PlaidItemsList.getPlaidItemsForDisplay(
+                    List<PlaidItem> items = PlaidItemsList.getPlaidItemsForDisplayExpanded(
                             adapter.getItems(), Collections.singletonList(newStory), columns);
                     adapter.setItems(items);
                     break;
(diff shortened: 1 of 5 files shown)


app/src/main/java/io/plaidapp/ui/HomeViewModel.kt
```diff
@@ -110,6 +106,14 @@ class HomeViewModel(
         loadData()
     }
 
+    fun getFeed(columns: Int): LiveData<FeedUiModel> {
+        return Transformations.switchMap(feedData) {
+            expandPopularItems(it, columns)
+            _feed.value = FeedUiModel(it)
+            _feed
```
nickbutcher: Think a labelled return would be clearer here?

app/src/test/java/io/plaidapp/ui/HomeViewModelTest.kt
```diff
@@ -281,7 +281,7 @@ class HomeViewModelTest {
         filtersChangedCallback.value.onFilterRemoved(dribbbleSource.key)
 
         // Then feed emits a new list, without the removed filter
-        val feed = LiveDataTestUtil.getValue(homeViewModel.feed)
+        val feed = LiveDataTestUtil.getValue(homeViewModel.getFeed(2))
```
nickbutcher: Factor out a `val columns = 2` for readability and to avoid magic numbers?

app/src/main/java/io/plaidapp/ui/HomeActivity.java
```diff
@@ -492,7 +489,7 @@ public void onReceive(Context context, Intent intent) {
                     // actually add the story to the grid
                     Story newStory = intent.getParcelableExtra(PostStoryService.EXTRA_NEW_STORY);
 
-                    List<PlaidItem> items = PlaidItemsList.getPlaidItemsForDisplay(
+                    List<PlaidItem> items = PlaidItemsList.getPlaidItemsForDisplayExpanded(
```
yigit: i was trying to figure out what is going on here, then noticed it is java :man_facepalming:

app/src/main/java/io/plaidapp/ui/HomeViewModel.kt
```diff
@@ -110,6 +106,14 @@ class HomeViewModel(
         loadData()
     }
 
+    fun getFeed(columns: Int): LiveData<FeedUiModel> {
+        return Transformations.switchMap(feedData) {
+            expandPopularItems(it, columns)
```
yigit: this should create a LiveData instead of keeping feed in the field. (get rid of _feed). (assuming feedData keeps the raw data).

yigit: once we have the lib, we should use buildLiveData here to offload the expansion to another dispatcher.

app/src/main/java/io/plaidapp/ui/HomeActivity.java
florina-muntenescu: 🤷

florina-muntenescu:
@yigit & @nickbutcher - ready for another review

app/src/main/java/io/plaidapp/ui/HomeViewModel.kt
```diff
@@ -109,8 +108,9 @@ class HomeViewModel(
     fun getFeed(columns: Int): LiveData<FeedUiModel> {
         return Transformations.switchMap(feedData) {
             expandPopularItems(it, columns)
-            _feed.value = FeedUiModel(it)
-            _feed
+            val feed = MutableLiveData<FeedUiModel>()
```
nickbutcher: Isn't there a constructor that accepts an initial value now? Did that launch yet?

app/src/main/java/io/plaidapp/ui/HomeViewModel.kt
```diff
@@ -110,6 +105,13 @@ class HomeViewModel(
         loadData()
     }
 
+    fun getFeed(columns: Int): LiveData<FeedUiModel> {
+        return Transformations.switchMap(feedData) {
+            expandPopularItems(it, columns)
```
nickbutcher: Drop a todo to move this work to the background?

PR #657, opened 2019-03-13
Remove DataLoadingSubject from InfiniteScrolling class

florina-muntenescu:


app/src/main/java/io/plaidapp/ui/HomeActivity.java
```diff
@@ -233,11 +234,20 @@ public int getSpanSize(int position) {
...
+                new InfiniteScrollListener(layoutManager) {
                     @Override
                     public void onLoadMore() {
                         viewModel.loadData();
                     }
+
+                    @Override
+                    public boolean isDataLoading() {
+                        FeedProgressUiModel uiModel = viewModel.getFeedProgress().getValue();
+                        if(uiModel !=  null){
```
nickbutcher:
```suggestion
                        if (uiModel !=  null) {
```

manuelvicnt: +1

Issue #655, opened 2019-03-13
Test two LiveData consequent emissions in ShotViewModel 


manuelvicnt:
There's a test missing in ShortViewModelTest.

We want to test that the ViewModel emits two uiModel objects. The first one as a fast emission as soon as the VM is created and then, after a new coroutine in the background makes some processing, the second value.

Maybe there's a way we can control the Coroutine with a `ContinuationInterceptor`? We don't want to change the VM behaviour just for the sake of the tests.

ashdavies:
Can you use `TestCoroutineContext.triggerActions()`?

manuelvicnt:
I tried this but doesn't work

```
    @UseExperimental(ObsoleteCoroutinesApi::class)
    @Test
    fun loadShot_emitsTwoUiModels() {
        val testContext = TestCoroutineContext()
        runBlocking(testContext) {
            // Given that the repo successfully returns the requested shot
            // When view model is constructed
            testContext.cancelAllActions()

            val viewModel = withViewModel()

            // Then a shotUiModel is present
            val slowResult: ShotUiModel? = LiveDataTestUtil.getValue(viewModel.shotUiModel)
            assertNotNull(slowResult)
            assertTrue(slowResult?.formattedDescription!!.isEmpty())

            testContext.triggerActions()

            // Then a shotUiModel is present
            val fastResult: ShotUiModel? = LiveDataTestUtil.getValue(viewModel.shotUiModel)
            assertNotNull(fastResult)
            assertTrue(slowResult?.formattedDescription!!.isNotEmpty())
        }
    }
```

Am I misusing it? should that work? @ashdavies

ashdavies:
This is how I would expect to use it, `cancelAllActions` may not be necessary, I'll try and have a go myself this evening, but if it's not working for you, perhaps it's not going to be the right solution 😢

ashdavies:
Looking over it again I think I can see what's wrong, it's not enough that you run the test code within the `TestCoroutineContext` (in fact I think that's unnecessary), the context needs to be provided to the `ViewModel` to ensure that the coroutine launched after the fast emission uses this context instead of the `Unconfined` that has been injected via `CoroutinesDispatcherProvider`, which would cause it to execute immediately.

[...]

PR #660, opened 2019-03-14
[DO NOT MERGE] Introduce Coroutine Context provider for ViewModel testability

ashdavies:
## DO NOT MERGE

In #655 we need to be able to control the execution of coroutines in a `ViewModel` in order to assert the expected "immediate" behaviour before the suspended operation would overwrite this value.

Since traditional test dispatchers provided defer to "immediate" execution using `Dispatchers.Unconfined`, this cannot be directly achieved.

This pull request, demonstrates how we can use a similar behaviour to provide a `CoroutineContext` instead of `CoroutineDispatcher` to test `ViewModel`'s behaviour, the test has been written by @manuelvicnt in #655, I've only demonstrated here how to use `TestCoroutineContext`.

Adversely, I would actually recommend against this solution, since it introduces confusion as to whether one should use `CoroutineContext` or `CoroutineDispatcher` and does not clearly identify the nuances between them.

Furthermore, a [pull request](https://github.com/Kotlin/kotlinx.coroutines/pull/890) from @objcode proposes a better implementation of `TestCoroutineContext`, splitting it up into `TestCoroutineDispatcher` and accompanying `TestCoroutineScope`, properly respecting the hierarchical nature of coroutine execution.

It would probably be better to wait for this to become available in `coroutine-test` rather than going with this implementation, but I have no idea how long that might be.

### Consideration
In our workflow we conside a pull request to be invalid if we don't provide a meme to accompany it, so here is mine 😄

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,10 @@
 buildscript { scriptHandler ->
     apply from: 'repositories.gradle',
             to: scriptHandler
-    ext.versions = [
+  repositories {
+    google()
+  }
+  ext.versions = [
             'compileSdk'         : 28,
             'minSdk'             : 23,
             'targetSdk'          : 28,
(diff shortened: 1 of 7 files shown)


manuelvicnt:
Thanks for the contribution Ash. This is a good understanding of how the Coroutine Context works but unfortunately we're going to wait until there are some improvements in this space to get that test in place. The changes to make that happen introduce some inconsistencies that we prefer to avoid. Thanks for your time on this, it helped us understand what's needed to have that working. Closing the PR and hoping we can create one to solve the issue soon :)

ashdavies:
I would absolutely agree that it's worth waiting for further improvements, especially with regard to structured concurrency, I'm glad this was helpful and I hopefully they won't be too long 😉

PR #667, opened 2019-03-20
Converting DataManager to Kotlin and removing coroutine launches from use cases

florina-muntenescu:


core/src/main/java/io/plaidapp/core/data/DataManager.kt
```diff
@@ -0,0 +1,258 @@
...
+    private lateinit var pageIndexes: MutableMap<String, Int>
+
+    private val filterListener = object : FiltersChangedCallback() {
+        override fun onFiltersChanged(changedFilter: SourceItem) {
+            if (changedFilter.active) {
+                loadSource(changedFilter)
+            } else { // filter deactivated
+                val key = changedFilter.key
+                parentJobs.filter { it.key.key == key }.forEach { job ->
+                    parentJobs.remove(job.key)
```
manuelvicnt: Apart from removing it, you have to cancel it, right? What do you think about doing a `cancelChildren()`?

nickbutcher: `cancelChildren` cancels *all* `Job`s, we just want to cancel those with the same source.

core/src/main/java/io/plaidapp/core/data/DataManager.kt
```diff
@@ -0,0 +1,255 @@
...
+                }
+                // clear the page index for the source
+                pageIndexes[key] = 0
+            }
+        }
+    }
+
+    init {
+        setOnDataLoadedCallback(onDataLoadedCallback)
```
nickbutcher: Isn't this just setting `null`? Seems redundant?

core/src/main/java/io/plaidapp/core/data/DataManager.kt
manuelvicnt: is `cancelChildren()` a thing with structured concurrency?

core/src/main/java/io/plaidapp/core/data/DataManager.kt
```diff
@@ -0,0 +1,255 @@
...
+                else -> if (source is DribbbleSourceItem) {
+                    parentJobs[data] = loadDribbbleSearch(source, data)
+                } else if (source is DesignerNewsSearchSource) {
+                    parentJobs[data] = loadDesignerNewsSearch(source, data)
+                }
+            }
+        }
+    }
+
+    private fun setupPageIndexes() {
```
nickbutcher: Wait, isn't this going to always fail the `if` check and not actually set up page indexes? Can we just remove this method and rely on `getNextPage` to populate `pageIndexes` on request?

core/src/main/java/io/plaidapp/core/data/DataManager.kt
```diff
@@ -0,0 +1,255 @@
...
+            if (pageIndexes[it.key] != null) {
+                pageIndexes[it.key] = 0
+            }
+        }
+    }
+
+    private fun getNextPageIndex(dataSource: String): Int {
+        var nextPage = 1 // default to one – i.e. for newly added sources
+        if (pageIndexes.containsKey(dataSource)) {
+            nextPage = pageIndexes[dataSource]?.or(0)?.plus(1) ?: 1
```
nickbutcher:
```suggestion
            nextPage = pageIndexes.getOrDefault(dataSource, 0) + 1
```

core/src/main/java/io/plaidapp/core/data/DataManager.kt
nickbutcher: Hmm, this might break `sourceIsEnabled` — is that working now?

core/src/main/java/io/plaidapp/core/data/DataManager.kt
florina-muntenescu: wooops. yeah. that's what I wanted

florina-muntenescu: Since we're also never using that Job again, I think we can just use `cancel` and not `cancelChildren`

core/src/main/java/io/plaidapp/core/data/DataManager.kt
```diff
@@ -0,0 +1,255 @@
...
+        // this API's paging is 0 based but this class (& sorting) is 1 based so adjust locally
+        val result = loadPosts(data.page - 1)
+        when (result) {
+            is Result.Success -> sourceLoaded(result.data, SOURCE_PRODUCT_HUNT, data)
+            is Result.Error -> loadFailed(data)
+        }.exhaustive
+    }
+
+    override fun registerCallback(callback: DataLoadingSubject.DataLoadingCallbacks) {
+        if (loadingCallbacks == null) {
```
nickbutcher: Use a `lazy` delegate?

core/src/main/java/io/plaidapp/core/data/DataManager.kt
```diff
@@ -0,0 +1,255 @@
...
+    }
+
+    private fun loadFinished() {
+        if (0 == loadingCount.decrementAndGet()) {
+            dispatchLoadingFinishedCallbacks()
+        }
+    }
+
+    private fun setPage(items: List<PlaidItem>, page: Int) {
+        for (item in items) {
```
nickbutcher: foreach

core/src/main/java/io/plaidapp/core/data/DataManager.kt
```diff
@@ -0,0 +1,255 @@
...
+            item.page = page
+        }
+    }
+
+    private fun setDataSource(items: List<PlaidItem>, dataSource: String) {
+        for (item in items) {
```
nickbutcher: foreach

core/src/main/java/io/plaidapp/core/data/DataManager.kt
```diff
@@ -0,0 +1,255 @@
...
+    suspend fun loadAllDataSources() {
+        sourcesRepository.getSources().forEach { loadSource(it) }
+    }
+
+    fun cancelLoading() {
+        parentJobs.values.forEach { it.cancel() }
+        parentJobs.clear()
+    }
+
+    private fun loadSource(source: SourceItem) {
```
nickbutcher: So this will ignore any scope used to invoke `loadAllDataSources` and `launch` with `scope` instead? Is that the desired behaviour here?

app/src/main/java/io/plaidapp/ui/HomeViewModel.kt
```diff
@@ -122,8 +122,8 @@ class HomeViewModel(
         designerNewsLoginRepository.logout()
     }
 
-    fun loadData() {
-        dataManager.loadAllDataSources()
+    fun loadData() = viewModelScope.launch(dispatcherProvider.computation) {
```
manuelvicnt: Since most interactions are with the View, the convention is that ViewModel always launch on `UI` and let the other classes decide about which threads to use. Said that, this can become `viewModelScope.launch { dataManager.loadMore() } ` since the dataManager should do the `withContext`.

Reasoning: DataManager has more information of what Dispatchers to use than the VM

core/src/main/java/io/plaidapp/core/data/DataManager.kt
```diff
@@ -0,0 +1,241 @@
...
+    private val loadStories: LoadStoriesUseCase,
+    private val loadPosts: LoadPostsUseCase,
+    private val searchStories: SearchStoriesUseCase,
+    private val shotsRepository: ShotsRepository,
+    private val sourcesRepository: SourcesRepository,
+    private val dispatcherProvider: CoroutinesDispatcherProvider
+) : DataLoadingSubject {
+
+    private var parentJob = SupervisorJob()
+    private val scope = CoroutineScope(dispatcherProvider.main + parentJob)
```
manuelvicnt: DataManager should never have to deal with the Main thread. Should this be `dispatcherProvider.default`?

core/src/main/java/io/plaidapp/core/data/DataManager.kt
```diff
@@ -0,0 +1,241 @@
...
+class DataManager @Inject constructor(
+    private val loadStories: LoadStoriesUseCase,
+    private val loadPosts: LoadPostsUseCase,
+    private val searchStories: SearchStoriesUseCase,
+    private val shotsRepository: ShotsRepository,
+    private val sourcesRepository: SourcesRepository,
+    private val dispatcherProvider: CoroutinesDispatcherProvider
+) : DataLoadingSubject {
+
+    private var parentJob = SupervisorJob()
```
manuelvicnt: I'd have `parentJob` as a `val`. The only reason I see for this being a `var` is if we wanted to cancel this job and recreate it

core/src/main/java/io/plaidapp/core/data/DataManager.kt
```diff
@@ -0,0 +1,241 @@
...
+                }
+                // clear the page index for the source
+                pageIndexes[key] = 0
+            }
+        }
+    }
+
+    init {
+        sourcesRepository.registerFilterChangedCallback(filterListener)
+        pageIndexes = sourcesRepository.getSourcesSync().map { it.key to 0 }.toMap().toMutableMap()
```
manuelvicnt: Don't understand what's happening here. Can you extract this to a method with a meaningful method name? Also, `toMap().toMutableMap()`? Can you do `.toMutableMap()` straight away?

core/src/main/java/io/plaidapp/core/data/DataManager.kt
```diff
@@ -0,0 +1,241 @@
...
+    }
+
+    fun cancelLoading() {
+        parentJobs.values.forEach { it.cancel() }
+        parentJobs.clear()
+    }
+
+    private fun loadSource(source: SourceItem) {
+        if (source.active) {
+            loadStarted()
```
manuelvicnt: This is being executed on the calling thread which in this case comes from `Dispatchers.computation` that you used in the VM. My comment above was about launching in `Dispatchers.Main`. Wouldn't it be better if this function (or even better `loadMore()`) is wrapped around a `withContext(Dispatchers.computation)`?

core/src/main/java/io/plaidapp/core/data/DataManager.kt
```diff
@@ -0,0 +1,241 @@
...
+        parentJobs.remove(request)
+    }
+
+    private fun loadFailed(request: InFlightRequestData) {
+        loadFinished()
+        parentJobs.remove(request)
+    }
+
+    private fun launchLoadDesignerNewsStories(data: InFlightRequestData) =
+        scope.launch(dispatcherProvider.computation) {
```
manuelvicnt: If we had `computation` as the default dispatcher, we wouldn't need this here

core/src/main/java/io/plaidapp/core/data/prefs/SourcesRepository.kt
```diff
@@ -40,8 +40,8 @@ class SourcesRepository(
         callbacks.add(callback)
     }
 
-    suspend fun getSources() = withContext(dispatcherProvider.io) {
-        getSourcesSync()
+    suspend fun getSources(): List<SourceItem> = withContext(dispatcherProvider.io) {
+        return@withContext getSourcesSync()
```
manuelvicnt: Do you need the `return@withContext`? doesn't it pick it up automatically?

core/src/main/java/io/plaidapp/core/data/DataManager.kt
florina-muntenescu: no, we can't do `.toMutableMap` :|

core/src/main/java/io/plaidapp/core/data/DataManager.kt
florina-muntenescu: no idea why this was `var`. good catch!

core/src/main/java/io/plaidapp/core/data/DataManager.kt
florina-muntenescu: comment added

core/src/main/java/io/plaidapp/core/data/prefs/SourcesRepository.kt
florina-muntenescu: nope.

core/src/main/java/io/plaidapp/core/data/DataManager.kt
florina-muntenescu: wrapping `loadMore` in `withContext(Dispatchers.computation)`

test_shared/build.gradle
```diff
@@ -45,6 +45,8 @@ dependencies {
     implementation "org.jetbrains.kotlin:kotlin-stdlib-jdk7:${versions.kotlin}"
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-core:${versions.coroutines}"
     implementation "org.jetbrains.kotlinx:kotlinx-coroutines-android:${versions.coroutines}"
+    implementation "junit:junit:${versions.junit}"
+    implementation "org.jetbrains.kotlinx:kotlinx-coroutines-test:${versions.coroutines}"
```
manuelvicnt: `testImplementation`?

florina-muntenescu: The utility classes we have here are not part of a testing module. They're the main files of the library, just that this library is only used for testing.

PR #676, opened 2019-03-26
Change designer news login theme to provide a dialog like UI experience

### LINKED ISSUE #444: [Bug] Strange background under Login view
When close Login (press back) view we see a semi-transparent black background under the view. Is that the commitment of designer?

Device: Samsung S8
Android:  Android 7.x

screen shot 2018-07-27 at 11 14 34

apoorv9990:
Fixes #444 

The `Plaid.Translucent.DesignerNewsLogin` style is currently deriving `android:windowBackground` attribute which is causing the shadow background. 

I changed the `android:windowBackground` to transparent and set `android:windowIsFloating` to true.

For some reason setting `android:windowIsFloating` to true has a weird effect on the width of the Activity.

screenshot-2019-03-25_18 43 50 095

so I set `android:windowMinWidthMajor` and `android:windowMinWidthMinor` explicitly.

### **Current Implementation**
Current Implementation

### **After Changes**
After changes

### DIFF
--- a/designernews/src/main/res/values/styles.xml
+++ b/designernews/src/main/res/values/styles.xml
@@ -22,6 +22,10 @@
         <item name="android:statusBarColor">@android:color/transparent</item>
         <item name="android:navigationBarColor">@android:color/transparent</item>
         <item name="android:windowAnimationStyle">@null</item>
+        <item name="android:windowBackground">@android:color/transparent</item>
+        <item name="android:windowIsFloating">true</item>
+        <item name="android:windowMinWidthMajor">100%</item>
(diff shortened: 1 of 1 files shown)


apoorv9990:
I created this PR but I have no idea why the CircleCI builds are failing. If someone could point me in the right direction on how to resolve it I would really appreciate it.

tiembo:
Hi @apoorv9990, thanks for the PR.  I've restarted the CI build.

nickbutcher:
Thank you for the PR! However as this seems to be a device specific issue (as `windowBackground` works elsewhere) and this change introduces more complexity into the solution I don't think that we should merge it.

PR #680, opened 2019-04-02
Splitting data sources for Search in different modules

florina-muntenescu:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description
Breaking data sources. Now each module implements an interface for the data source. app and search modules don't know about the concrete implementation of the data sources
Using Dagger and reflection to add the search data sources in a registry of searchable data sources

The concrete implementation of data sources and data retrieval for each data source shouldn't be visible outside the module that implements that data source. 

Manual + unit tests

## :pencil: Checklist

- [x] I added tests to verify changes
- [x] All tests passing


search/src/main/java/io/plaidapp/search/ui/SearchActivity.kt
```diff
@@ -109,7 +110,9 @@ class SearchActivity : AppCompatActivity() {
 
         feedAdapter = FeedAdapter(this, columns, pocketInstalled)
 
-        viewModel.searchResults.observe(this, EventObserver { searchUiModel ->
+        sendBroadcast(Intent("io.plaidapp.register.SEARCH_FACTORY"))
```
wojtek-kalicinski: this doesn't work for me because of implicit intent + background restriction
you can add a package name to make it explicit:
`sendBroadcast(Intent("io.plaidapp.register.SEARCH_FACTORY").setPackage(packageName))`

and then you should get rid of all the services, they're not needed. you can do the provider registration in the receiver itself

dribbble/src/main/java/io/plaidapp/dribbble/dagger/DribbbleSearchComponent.kt
```diff
@@ -0,0 +1,35 @@
...
+import io.plaidapp.core.dagger.dribbble.DribbbleDataModule
+import io.plaidapp.core.dagger.scope.FeatureScope
+import io.plaidapp.core.interfaces.SearchDataSourceFactory
+
+@Component(
+    modules = [
+        DribbbleDataModule::class,
+        SearchDataModule::class
+    ],
+    dependencies = [io.plaidapp.core.dagger.CoreComponent::class]
```
manuelvicnt: import `io.plaidapp.core.dagger`

search/src/main/java/io/plaidapp/search/dagger/SearchModule.kt
```diff
@@ -24,24 +24,17 @@ import dagger.Binds
...
 
-@Module(
-    includes = [
-        DribbbleDataModule::class,
-        DesignerNewsDataModule::class,
-        SharedPreferencesModule::class
-    ]
-)
+@Module
 abstract class SearchModule {
```
manuelvicnt: Why is this abstract and then contains a `companion object` inside it?

manuelvicnt: Wouldn't need the `@JvmStatic` either as the Component is also a Kotlin file

search/src/main/java/io/plaidapp/search/domain/LoadSearchDataUseCase.kt
```diff
@@ -0,0 +1,52 @@
...
+) {
+
+    private val dataSources = factories.map { it.create(query) }
+
+    private val _searchResult = MutableLiveData<List<PlaidItem>>()
+    val searchResult: LiveData<List<PlaidItem>>
+        get() = _searchResult
+
+    suspend operator fun invoke() {
+        dataSources.forEach {
```
manuelvicnt: Does this need to be run sequentially? We could use `async` to make these requests in parallel and wait for all of them to finish :)

search/src/main/java/io/plaidapp/search/domain/SearchDataSourceFactoriesRegistry.kt
```diff
@@ -0,0 +1,45 @@
...
+    private val _dataSourceFactories = mutableSetOf<SearchDataSourceFactory>()
+
+    init {
+        defaultFactories.get()?.apply { _dataSourceFactories.addAll(this) }
+    }
+
+    val dataSourceFactories: Set<SearchDataSourceFactory>
+        get() = _dataSourceFactories
+
+    fun add(dataSourceFactory: SearchDataSourceFactory) {
```
manuelvicnt: add and remove don't seem to be used

search/src/main/java/io/plaidapp/search/ui/SearchViewModel.kt
```diff
@@ -18,66 +18,56 @@ package io.plaidapp.search.ui
...
     }
 
-    fun searchFor(query: String) = viewModelScope.launch(dispatcherProvider.computation) {
-        dataManager.searchFor(query)
+    private val _searchProgress = MutableLiveData<FeedProgressUiModel>()
+    val searchProgress: LiveData<FeedProgressUiModel>
+        get() = _searchProgress
+
+    fun searchFor(query: String) {
+        searchQuery.postValue(query)
```
manuelvicnt: This is supposed to be run from the main thread, right? Do we have any convention of using `postValue` always?

search/src/main/java/io/plaidapp/search/ui/SearchViewModel.kt
```diff
@@ -18,66 +18,56 @@ package io.plaidapp.search.ui
...
-    init {
-        dataManager.onDataLoadedCallback = onDataLoadedCallback
-        dataManager.registerCallback(dataLoadingCallbacks)
+    val searchResults: LiveData<FeedUiModel> = Transformations.map(results) {
+        FeedUiModel(it)
     }
 
-    fun searchFor(query: String) = viewModelScope.launch(dispatcherProvider.computation) {
-        dataManager.searchFor(query)
+    private val _searchProgress = MutableLiveData<FeedProgressUiModel>()
```
manuelvicnt: Is there any particular reason we're exposing two UI LiveData objects? Can it not be just one? 

```
data class FeedUiModel(
    val items: List<PlaidItem>,
    val isLoading: Boolean
)
```

core/src/main/java/io/plaidapp/core/producthunt/data/api/model/Post.kt
```diff
@@ -16,14 +16,11 @@
...
 package io.plaidapp.core.producthunt.data.api.model
 
-import android.os.Parcelable
 import io.plaidapp.core.data.PlaidItem
-import kotlinx.android.parcel.Parcelize
 
 /**
  * Models a post on Product Hunt.
  */
-@Parcelize
```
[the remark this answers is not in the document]
florina-muntenescu: No. it's not really passed anywhere anymore.

core/src/main/java/io/plaidapp/core/interfaces/PlaidDataSource.kt
```diff
@@ -0,0 +1,33 @@
...
+
+import io.plaidapp.core.data.PlaidItem
+import io.plaidapp.core.data.Result
+import io.plaidapp.core.data.SourceItem
+
+/**
+ * Class that allows loading data based on a [SourceItem].
+ * Implementations of this class will handle how the data is retrieved and how pages are handled
+ */
+abstract class PlaidDataSource(val sourceItem: SourceItem) {
```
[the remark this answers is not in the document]
florina-muntenescu: Indeed, they both follow the same pattern, but I feel like adding abstract methods for getting data from a repo AND for parsing the success response is an overkill for how small these implementations are.

designernews/src/main/java/io/plaidapp/designernews/dagger/DesignerNewsSearchComponent.kt
```diff
@@ -0,0 +1,38 @@
...
+
+@Component(
+    modules = [
+        CoreDataModule::class,
+        DataModule::class,
+        DesignerNewsDataModule::class,
+        SearchDataModule::class,
+        DesignerNewsPreferencesModule::class
+    ]
+)
```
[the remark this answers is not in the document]
florina-muntenescu: Why? Can you give more details?

search/src/main/java/io/plaidapp/search/dagger/SearchModule.kt
florina-muntenescu: Should then this be an interface?
If I make it an interface and remove `@JvmStatic` I get
```
error: [Dagger/MissingBinding] java.util.Set<io.plaidapp.core.interfaces.SearchDataSourceFactory> cannot be provided without an @Provides-annotated method.
    public abstract java.util.Set<io.plaidapp.core.interfaces.SearchDataSourceFactory> factories();
                                                                                       ^
      java.util.Set<io.plaidapp.core.interfaces.SearchDataSourceFactory> is provided at
          io.plaidapp.search.dagger.SearchComponent.factories()
```

search/src/main/java/io/plaidapp/search/ui/SearchViewModel.kt
florina-muntenescu: I don't think we have any conventions, but I'll `set` this.

search/src/main/java/io/plaidapp/search/ui/SearchViewModel.kt
florina-muntenescu: There's more stuff I'd want to do here with the `FeedUiModel`. 
I'll leave this for a different PR.

PR #685, opened 2019-04-13
Using dagger to combine data sources for Search that were split into different modules

florina-muntenescu:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
Using Dagger instead of `BroadcastReceiver` to combine data sources

Looking for another solution instead of `BroadcastReceiver`s

Manual testing

## :pencil: Checklist

- [x] All tests passing

## :crystal_ball: Next steps
Merging this into #680


search/src/main/java/io/plaidapp/search/dagger/SearchModule.kt
```diff
@@ -61,6 +56,37 @@ abstract class SearchModule {
...
+        }
+
+        private fun searchDataSourceFactory(
+            activity: Activity,
+            className: String
+        ): SearchDataSourceFactory? {
+            return try {
+                val provider =
+                    Class.forName(className).kotlin.objectInstance as SearchFactoryProvider
+                provider.getFactory(activity.applicationContext)
```
manuelvicnt: Use regular context. We should always use activity context and only use applicationContext for non-visual things like Services.

designernews/src/main/java/io/plaidapp/designernews/domain/search/DesignerNewsSearchDataSourceFactoryProvider.kt
```diff
@@ -18,25 +18,17 @@ package io.plaidapp.designernews.domain.search
...
 import android.content.Context
 import io.plaidapp.core.interfaces.SearchDataSourceFactory
-import io.plaidapp.core.interfaces.SearchFactoryProvider
+import io.plaidapp.core.interfaces.SearchDataSourceFactoryProvider
 import io.plaidapp.designernews.dagger.DaggerDesignerNewsSearchComponent
 import io.plaidapp.designernews.dagger.DesignerNewsPreferencesModule
-import javax.inject.Inject
 
-class DesignerNewsSearchFactoryProvider(context: Context) : SearchFactoryProvider {
+object DesignerNewsSearchDataSourceFactoryProvider : SearchDataSourceFactoryProvider {
```
manuelvicnt: Why an `object`? May difficult testing

dribbble/src/main/java/io/plaidapp/dribbble/dagger/DribbbleSearchComponent.kt
```diff
@@ -0,0 +1,36 @@
...
+    modules = [
+        DribbbleDataModule::class,
+        SearchDataModule::class
+    ],
+    dependencies = [io.plaidapp.core.dagger.CoreComponent::class]
+)
+@FeatureScope
+interface DribbbleSearchComponent {
+
+    @FeatureScope
```
manuelvicnt: The annotation is not needed here. It's ultimately taken from the one in the `@Provides` method

search/src/main/java/io/plaidapp/search/dagger/SearchModule.kt
```diff
@@ -61,6 +58,38 @@ abstract class SearchModule {
         @Provides
         fun isPocketInstalled(activity: Activity): Boolean = PocketUtils.isPocketInstalled(activity)
 
+        @JvmStatic
+        @Provides
+        @FeatureScope
+        fun factories(activity: Activity): Set<SearchDataSourceFactory> {
```
ZacSweers: The reflection isn't terrible, but the hardcoded names seems error prone. I think one (equally awkward, but safer for refactoring) way could be to create a default method in the interface that registers itself to some global static registry that can be read from here at runtime. Unsure if you'd potentially be racing the classloader, but that'd let you avoid the class name hardcoding.

```kotlin
interface SearchDataSourceFactoryProvider {
  fun register() {
    REGISTERED_IMPLEMENTATIONS.add(this) // or javaClass if you're not implementing these only in singleton `object`s and want to reflectively invoke a new instance
  }
  
  companion object Registry {
    val REGISTERED_IMPLEMENTATIONS: MutableSet<SearchDataSourceFactoryProvider> = ConcurrentHashMap< SearchDataSourceFactoryProvider, Boolean>().keySet() // Resolve a concurrent set
  }
}
```

PR #693, opened 2019-05-01
Deduplicate configuration of repositories in build.gradle files

egor-n:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

The following line has been extracted from all `build.gradle` files into the `subprojects {}` block.

```
apply from: '../repositories.gradle'
```

I was looking through #615 and spotted that all `build.gradle` files have the `apply from: 'repositories.gradle'` line. This change is just reducing the clutter in each of the `build.gradle` files.

## :pencil: Checklist

- [x] All tests passing

## :crystal_ball: Next steps
1. All modules have similar `android { defaultConfig }` configuration which could also be deduplicated and extracted into the `subprojects {}` block. Let me know if this is something worth doing.
2. `apply plugin: 'kotlin-android'` can also be extracted to `subprojects {}` but that would bring issues if some non-Android modules are planned.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -82,6 +82,10 @@ ext {
 }
 
 subprojects {
+    buildscript {
+        apply from: rootProject.file('repositories.gradle')
+    }
+
     apply plugin: 'com.diffplug.gradle.spotless'
     spotless {
         kotlin {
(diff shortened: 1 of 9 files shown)


build.gradle
```diff
@@ -82,6 +82,10 @@ ext {
 }
 
 subprojects {
+    buildscript {
+        apply from: rootProject.file('repositories.gradle')
```
piotrek1543: I am not maintainer of this repo, but in the same way you could reduce re-declaration of `core_dependencies.gradle` and `test_dependencies.gradle`. Anyway, I like your point of view :-)

PR #694, opened 2019-05-13
[DO NOT MERGE] Adds TestCoroutineDispatcher + LiveData coroutines support

### LINKED ISSUE #655: Test two LiveData consequent emissions in ShotViewModel

manuelvicnt:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description

Used the new experimental TestCoroutineDispatcher to fix #655.

Before, we didn't have the option to stop / resume the CoroutinesDispatcher when injected with our `CoroutinesDispatcherProvider` so we couldn't test logic that relies on that.

Unit Tests

## :pencil: Checklist

- [x] All tests passing

## :crystal_ball: Next steps
When the API hits stable, we'll migrate all the tests to use this. `runBlocking` in tests should migrate to `runBlockingTest`.

### DIFF
--- a/dribbble/src/test/java/io/plaidapp/dribbble/ui/shot/ShotViewModelTest.kt
+++ b/dribbble/src/test/java/io/plaidapp/dribbble/ui/shot/ShotViewModelTest.kt
@@ -32,15 +32,20 @@ import io.plaidapp.dribbble.testShot
 import io.plaidapp.dribbble.testShotUiModel
 import io.plaidapp.test.shared.LiveDataTestUtil
 import io.plaidapp.test.shared.provideFakeCoroutinesDispatcherProvider
+import kotlinx.coroutines.ExperimentalCoroutinesApi
 import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.test.TestCoroutineDispatcher
+import kotlinx.coroutines.test.runBlockingTest
 import org.junit.Assert.assertEquals
 import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertTrue
 import org.junit.Rule
 import org.junit.Test
 
 /**
  * Tests for [ShotViewModel], mocking out its dependencies.
  */
+@ExperimentalCoroutinesApi
 class ShotViewModelTest {
 
     // Executes tasks in the Architecture Components in the same thread
@@ -53,6 +58,7 @@ class ShotViewModelTest {
     private val createShotUiModel: CreateShotUiModelUseCase = mock {
         on { runBlocking { invoke(any()) } } doReturn testShotUiModel
     }
+    private val testCoroutineDispatcher = TestCoroutineDispatcher()
 
     @Test
     fun loadShot_existsInRepo() {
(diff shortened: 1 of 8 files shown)


dribbble/src/test/java/io/plaidapp/dribbble/ui/shot/ShotViewModelTest.kt
```diff
@@ -143,6 +149,26 @@ class ShotViewModelTest {
...
     }
 
+    @Test
+    fun loadShot_emitsTwoUiModels() {
+        // Given coroutines have not started yet and the View Model is created
+        testCoroutineDispatcher.pauseDispatcher()
+        val viewModel = withViewModel()
+
+        // Then the fast result has been emitted
+        val fastResult: ShotUiModel? = LiveDataTestUtil.getValue(viewModel.shotUiModel)
```
JoseAlcerreca: Is viewModel.shotUiModel nullable? Probably not a good thing but it's another story

test_shared/src/main/java/io/plaidapp/test/shared/FakeAppInjection.kt
```diff
@@ -17,7 +17,12 @@
 package io.plaidapp.test.shared
 
 import io.plaidapp.core.data.CoroutinesDispatcherProvider
+import kotlinx.coroutines.CoroutineDispatcher
 import kotlinx.coroutines.Dispatchers.Unconfined
 
-fun provideFakeCoroutinesDispatcherProvider(): CoroutinesDispatcherProvider =
-    CoroutinesDispatcherProvider(Unconfined, Unconfined, Unconfined)
+fun provideFakeCoroutinesDispatcherProvider(
+    main: CoroutineDispatcher = Unconfined,
```
JoseAlcerreca: I think you should document why you're using Unconfined here as a default.

JoseAlcerreca:
LGTM but "Before, we didn't have the option to stop / resume the CoroutinesDispatcher so we couldn't test logic that relies on that." is not completely true: 

https://github.com/JoseAlcerreca/android-testing-codelab/commit/7d15375faa8906f99fb95588fc65bf5fbcfe4e34#diff-1350fae97992e3de86aff3b066423618L79

dribbble/src/test/java/io/plaidapp/dribbble/ui/shot/ShotViewModelTest.kt
manuelvicnt: `LiveDataTestUtil.getValue` is what returns a nullable value :thinking:

test_shared/src/main/java/io/plaidapp/test/shared/FakeAppInjection.kt
manuelvicnt: Done! Thanks

manuelvicnt:
This PR needs to wait until the next coroutines release since the problem with the CI is already fixed in coroutines `develop` with this PR https://github.com/Kotlin/kotlinx.coroutines/pull/1160

test_shared/src/main/java/io/plaidapp/test/shared/FakeAppInjection.kt
```diff
@@ -17,7 +17,15 @@
...
 import kotlinx.coroutines.Dispatchers.Unconfined
 
-fun provideFakeCoroutinesDispatcherProvider(): CoroutinesDispatcherProvider =
-    CoroutinesDispatcherProvider(Unconfined, Unconfined, Unconfined)
+// Using Unconfined here as a way to execute coroutines in tests in the test thread.
+// That makes the test to wait for the coroutine to finish before carry on.
+// This needs to be improved in the future, either using a CoroutinesTestRule
+// or the recently added TestCoroutineDispatcher when it hits stable.
+fun provideFakeCoroutinesDispatcherProvider(
+    main: CoroutineDispatcher = Unconfined,
```
manuelvicnt: This needs to change. Discussing with Sean whether this should be multiple instances of `TestCoroutineDispatcher` (one instance per dispatcher) or a shared instance. My gut tells me this should be unique per dispatcher since there are scenarios where we launch multiple coroutines in different dispatchers at the same time. e.g. `HomeViewModel` init

PR #702, opened 2019-05-23
Fixes Dribbble description animation

### LINKED ISSUE #675: Dribbble description doesn't slide up with the rest of the content

manuelvicnt:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Fixes Dribbble description animation.

Fixes #675 

For some reason, the transition doesn't like the BindingAdapter `goneUnless`. Even if we modify the adapter to only set the new visibility when it changes, the animation stops working. 

My gut tells me that the view is always GONE because we emit a first ShotUiModel with an empty description and then when the description is loaded, we emit another model with description making the animation for that view to stop working. 

That's why the decision is to set the visibility to GONE in the htmlText BindingAdapter when the description is null.

Manually.

## :pencil: Checklist

- [x] All tests passing

## :crystal_ball: Next steps
None.

## :camera_flash: Screenshots / GIFs

fix

### DIFF
--- a/dribbble/src/main/java/io/plaidapp/dribbble/ui/shot/ShotBindingAdapters.kt
+++ b/dribbble/src/main/java/io/plaidapp/dribbble/ui/shot/ShotBindingAdapters.kt
@@ -18,6 +18,7 @@ package io.plaidapp.dribbble.ui.shot
 
 import android.graphics.drawable.AnimatedVectorDrawable
 import android.text.format.DateUtils
+import android.view.View
 import android.widget.ImageView
 import android.widget.TextView
 import androidx.annotation.PluralsRes
@@ -67,7 +68,10 @@ fun bindHtmlText(
     textView: TextView,
     htmlText: CharSequence?
 ) {
-    if (htmlText == null) return
+    if (htmlText == null) {
+        textView.visibility = View.GONE
+        return
+    }
     HtmlUtils.setTextWithNiceLinks(textView, htmlText)
 }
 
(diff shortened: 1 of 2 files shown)


dribbble/src/main/java/io/plaidapp/dribbble/ui/shot/ShotBindingAdapters.kt
```diff
@@ -67,7 +68,10 @@ fun bindHtmlText(
     textView: TextView,
```
nickbutcher: Rename attr & fun to indicate this new behavior? e.g. `htmlTextOrGone` & `bindHtmlTextOrGone`.

dribbble/src/main/java/io/plaidapp/dribbble/ui/shot/ShotBindingAdapters.kt
```diff
@@ -67,7 +68,10 @@ fun bindHtmlText(
...
     htmlText: CharSequence?
 ) {
-    if (htmlText == null) return
+    if (htmlText == null) {
+        textView.visibility = View.GONE
```
nickbutcher: Import method as elsewhere for readability?

dribbble/src/main/java/io/plaidapp/dribbble/ui/shot/ShotBindingAdapters.kt
manuelvicnt: 👍

dribbble/src/main/java/io/plaidapp/dribbble/ui/shot/ShotBindingAdapters.kt
manuelvicnt: Done! Thanks

PR #704, opened 2019-05-23
Convert HomeActivity.java -> HomeActivity.kt

keyboardsurfer:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

Convert HomeActivity to Kotlin.

Continue migrating major parts of Plaid to Kotlin.

Manually, then ran instrumentation tests, then waited for circle to complete.

## :pencil: Checklist

- [x] All tests passing

## :crystal_ball: Next steps

Add in app update API.

Figure out why `dexDebug` fails here as well as on `master` where `master` is passing and this branch is not.


nickbutcher:
As this is a large change, can we restrict it to `HomeActivity` only and revert all other files?

keyboardsurfer:
> As this is a large change, can we restrict it to `HomeActivity` only and revert all other files?

Done

app/src/main/java/io/plaidapp/ui/HomeActivity.kt
```diff
@@ -0,0 +1,850 @@
...
+    }
+
+    private fun animateToolbar() {
+        // this is gross but toolbar doesn't expose it's children to animate them :(
+        val t = toolbar.getChildAt(0)
+        if (t != null && t is TextView) {
+
+            // fade in and space out the title.  Animating the letterSpacing performs horribly so
+            // fake it by setting the desired letterSpacing then animating the scaleX ¯\_(ツ)_/¯
+            t.alpha = 0f
```
piotrek1543: Hi, just a small thing - what about using the same operator (I mean `t.apply{ }`) like in method `showFab()` below? It will reduce number of  `t` variable calls  and what is more both code snippets looks pretty similar to me :-)

keyboardsurfer: Good spot. Updated that section.

app/src/main/java/io/plaidapp/ui/HomeActivity.kt
```diff
@@ -0,0 +1,857 @@
...
+
+        val callback = FilterTouchHelperCallback(filtersAdapter, this)
+        ItemTouchHelper(callback).also {
+            it.attachToRecyclerView(filtersList)
+        }
+        checkEmptyState()
+    }
+
+    private fun initViewModelObservers() {
+        viewModel.apply {
```
florina-muntenescu: i don't really like this `apply`

all the methods called here are so big, that it doesn't feel like it makes sense to have the `apply` just to not use `viewModel`.
Plus, here you're not configuring the viewModel, so it should be a `let`. See [cheatsheet](https://medium.com/androiddevelopers/kotlin-standard-functions-cheat-sheet-27f032dd4326)

app/src/main/java/io/plaidapp/ui/HomeActivity.kt
```diff
@@ -0,0 +1,857 @@
...
+        fab = findViewById(R.id.fab)
+        fab.setOnClickListener { fabClick() }
+        filtersList = findViewById(R.id.filters)
+        loading = findViewById(android.R.id.empty)
+
+        columns = resources.getInteger(R.integer.num_columns)
+    }
+
+    private fun setupGrid() {
+        with(grid) {
```
florina-muntenescu: I feel like the `with` here makes the code less readable, since you don't know which fields and functions apply to the grid and which not. 
Consider removing the `with`

app/src/main/java/io/plaidapp/ui/HomeActivity.kt
```diff
@@ -0,0 +1,857 @@
...
+                // delay to allow the toggle anim to run
+                postDelayed({
+                    AppCompatDelegate.setDefaultNightMode(
+                        if (isChecked)
+                            AppCompatDelegate.MODE_NIGHT_YES
+                        else
+                            AppCompatDelegate.MODE_NIGHT_NO
+                    )
+                    delegate.applyDayNight()
+                }, 800L)
```
florina-muntenescu: formatting

app/src/main/java/io/plaidapp/ui/HomeActivity.kt
keyboardsurfer: :ok_hand:

app/src/main/java/io/plaidapp/ui/HomeActivity.kt
keyboardsurfer: :ok_hand:

app/src/main/java/io/plaidapp/ui/HomeActivity.kt
keyboardsurfer: :ok_hand:

PR #751, opened 2019-07-10
Use springs for RecyclerView.ItemAnimator

nickbutcher:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :scroll: Description
This change updates `SlideInItemAnimator` which is the base class of all custom `RecyclerView.ItemAnimator`s, to use `SpringAnimation`s when animating adding or moving items.

This change supports smoother, re-targetable animations. Now if an item moves to a new position while a move is already happening, it respects the velocity of the running animation, smoothly re-targeting to the new end position.

Manually

## :pencil: Checklist

- [x] All tests passing

## :crystal_ball: Next steps
None


core/src/main/java/io/plaidapp/core/util/SpringUtils.kt
```diff
@@ -0,0 +1,114 @@
...
+    stiffness: Float = 500f,
+    damping: Float = SpringForce.DAMPING_RATIO_NO_BOUNCY,
+    startVelocity: Float? = null
+): SpringAnimation {
+    val key = getKey(property)
+    var springAnim = getTag(key) as? SpringAnimation?
+    if (springAnim == null) {
+        springAnim = SpringAnimation(this, property).apply {
+            spring = SpringForce().apply {
+                this.dampingRatio = damping
```
keyboardsurfer: `this` can be confusing here. Consider naming the parameters.

nickbutcher: Actually, should probably always configure the spring damping/stiffness, even if reusing an existing SpringAnimator. Will update.

PR #759, opened 2019-08-05
Use api instead of implementation

RicardoBravoA:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [X] Refactoring

## :scroll: Description

`api` is being used instead of `implementation` in gradle


RicardoBravoA:
@googlebot I signed it!

emartynov:
Why would I want to use `api` instead of `implementation`? What is the motivation?

RicardoBravoA:
> Why would I want to use `api` instead of `implementation`? What is the motivation?

Hello, because in a modular architecture we have many modules that depend on other modules and having to use 'implementation' of 'n' modules is repetitive when we can use 'api' and have a single module as dependency, it helps too much especially when we have projects With many modules.

I comment from my own experience, I used this repository as the basis for the project I am working on, I currently have more than 30 modules and using 'api' has saved me many dependency lines.

emartynov:
Thank you! It is preferred to use `implementation` instead of `api` (one of the explanations https://www.folio3.com/blog/gradle-when-to-use-implementation-vs-api/) to avoid module recompilation in case of dependency change. Did you try to measure incremental compilation times with these changes?

keyboardsurfer:
In this case I'd rather keep it with `implementation`.
Using `api` removes the need to keep track of dependencies explicitly.
With [feature-on-feature](https://developer.android.com/studio/preview/features#feature-on-feature) dependencies it's even more important to be explicit here.

Issue #767, opened 2019-09-11
ConcurrentModificationException when ending animation


tunjid:
https://github.com/android/plaid/blob/1115fb60bbc870f296efe50471ad4746873f199e/core/src/main/java/io/plaidapp/core/ui/recyclerview/SlideInItemAnimator.kt#L183

In the endAnimations method, while in a for each loop, the operator function remove is called which results in a ConcurrentModificationException :(

PR #768, opened 2019-09-11
Added modifiableForEach extension method that allows for in place ListModification during traversal

tunjid:

- [x] Bugfix
- [ ] New feature
- [ ] Enhancement
- [ ] Refactoring

## :scroll: Description

Added modifiableForEach extension method that allows for in place List modification during traversal.

Adresses #767 

If the SlideInItemAnimator is running during an orientation change, a `ConcurrentModificationException` will be thrown. This change allows for traversal with an iterator, allowing for in place modifications to the list.

I used the animator in another project where the SlideInItemAnimator runs continuously,  and triggered Activity destruction by rotating the phone. Source for that class is here: https://github.com/tunjid/android-bootstrap/blob/develop/app/src/main/java/com/tunjid/androidbootstrap/fragments/ShiftingTileFragment.kt

## :pencil: Checklist

- [x] All tests passing

### DIFF
--- a/core/src/main/java/io/plaidapp/core/util/Extensions.kt
+++ b/core/src/main/java/io/plaidapp/core/util/Extensions.kt
@@ -39,7 +39,7 @@ val <T> T.exhaustive: T
  * A for each loop where traversal is backed by an [MutableIterator] allowing for removal
  * during iteration.
  */
-inline fun <T> MutableIterable<T>.removableForEach(action: (T) -> Unit) {
+inline fun <T> MutableIterable<T>.onEach(action: (T) -> Unit) {
     val iterator = iterator()
     while (iterator.hasNext()) action(iterator.next())
 }
--- a/core/src/main/java/io/plaidapp/core/util/Extensions.kt
+++ b/core/src/main/java/io/plaidapp/core/util/Extensions.kt
@@ -34,3 +34,13 @@ package io.plaidapp.core.util
  */
 val <T> T.exhaustive: T
     get() = this
+
+/**
+ * A for each loop where traversal is backed by an [Iterator] allowing for list modification
+ * in place.
+ */
+inline fun <T> Iterable<T>.modifiableForEach(action: (T) -> Unit) {
+    iterator().apply {
+        while (hasNext()) next().apply(action)
+    }
+}
--- a/core/src/main/java/io/plaidapp/core/util/Extensions.kt
+++ b/core/src/main/java/io/plaidapp/core/util/Extensions.kt
@@ -36,11 +36,10 @@ val <T> T.exhaustive: T
     get() = this
 
 /**
- * A for each loop where traversal is backed by an [Iterator] allowing for list modification
- * in place.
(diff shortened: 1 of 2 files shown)


keyboardsurfer:
Over to Nick since he's been working in the Animator.

core/src/main/java/io/plaidapp/core/util/Extensions.kt
```diff
@@ -34,3 +34,13 @@ package io.plaidapp.core.util
  */
 val <T> T.exhaustive: T
     get() = this
+
+/**
+ * A for each loop where traversal is backed by an [Iterator] allowing for list modification
+ * in place.
+ */
+inline fun <T> Iterable<T>.modifiableForEach(action: (T) -> Unit) {
```
nickbutcher: Not sure about the name here, to me 'modifiable', implies that this modifies each element, not that the underling collection can be modified.

core/src/main/java/io/plaidapp/core/util/Extensions.kt
```diff
@@ -34,3 +34,13 @@ package io.plaidapp.core.util
...
+    iterator().apply {
```
nickbutcher: The `apply` block feels less idiomatic here. Why not just `val iterator = iterator()` like [most of the stdlib does](https://github.com/JetBrains/kotlin/blob/master/libraries/stdlib/common/src/generated/_Collections.kt)?

core/src/main/java/io/plaidapp/core/util/Extensions.kt
```diff
@@ -34,3 +34,13 @@ package io.plaidapp.core.util
...
+        while (hasNext()) next().apply(action)
```
nickbutcher: again the `apply` feels unnecessary, why not just `while (iterator.hasNext()) next().action(it)`

core/src/main/java/io/plaidapp/core/util/Extensions.kt
```diff
@@ -34,3 +34,13 @@ package io.plaidapp.core.util
...
+    }
+}
```
nickbutcher: nit: newline

nickbutcher (changes requested):
Thanks for the contribution! Added a few review comments.

tunjid:
@nickbutcher thanks for the comments! Are rebases preferred to merges when there are changes upstream? I checked the contributing doc, but it doesn't seem to mention it.

nickbutcher:
@tunjid rebase please. Thanks!

tunjid:
I'm not sure of a better name than `removableForEach`, very much open to suggestions. I also narrowed the type on the extension to a `MutableIterable` to better communicate intent.

nickbutcher:
Hmm, I can't think of a great name either and worry that `removableForEach` implies that the action may remove an item. What about `onEach`? It's similar enough to `forEach` to convey what it does without suggesting extra behavior?

tunjid:
That works! What do you think about adding an iterator prefix so the mutable iterator backing is communicated? So `iteratorOnEach`?

PR #770, opened 2019-10-08
DO NOT MERGE - live data & coroutines experiments

yigit:

- [ ] Bugfix
- [ ] New feature
- [ ] Enhancement
- [x] Refactoring

## :scroll: Description

This PR has two parts, just sending to share some ideas.

First part is about using the liveData builder instead of launching in scopes and calling setValue on a LiveData instance.
This allows tests to just use LiveData.asFlow() instead of LiveDataTestUtils. Also much nicer organization on how the value is computed (since there is only 1 place to look)
For the test case we want 2 events, unfortunately events get conflated with LiveData.asFlow and test coroutine dispatcher executes all runnables immediately (instead of first come first serve) but adding a yield between two emits in the ViewModel solves it (not great).

The second part is replacing Event class with a regular kotlin channel.
Channels have a fan-out functionality which is explicit (unlike in SingleLiveEvent which had implicit behavior that does not match LiveData). I find it much cleaner and has room to be generalized.

I don't think it makes sense to merge this PR (hence DO NOT MERGE), but i think there is room to move more code into coroutines / flows also cleanup this PRs code.
I think even the LiveData computations can be reduced into Flows all the way until UI (viewModel's public fields that is). 

Avoiding LiveDataTestUtil.

ShotViewModelTest

## :pencil: Checklist

- [] I added tests to verify changes
- [x] All tests passing

## :crystal_ball: Next steps
just check it

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -1,4 +1,4 @@
-/*
+ /*
  * Copyright 2015 Google Inc.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
--- a/build.gradle
+++ b/build.gradle
@@ -42,7 +42,7 @@ buildscript { scriptHandler ->
             'kotlin'             : '1.3.50',
             'ktlint'             : '0.24.0',
             'legacyCoreUtils'    : '1.0.0',
-            'lifecycle'          : '2.2.0-alpha03',
+            'lifecycle'          : '2.2.0-alpha05',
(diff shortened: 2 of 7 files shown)


app/build.gradle
```diff
@@ -1,4 +1,4 @@
-/*
+ /*
```
nic0lette: super nit: It looks like a space got added here accidentally. :)

Issue #745, opened 2019-07-01
Review/FixSuppress all lint findings


emartynov:
It is quite unfortunate to see all lint findings in PR CI build log and guess what are new because of your recent changes.

Proposal to review all existing and after making any lint finding strict so it will break the build.

keyboardsurfer:
Thanks for the proposal and implementing changes in PRs.

PR #779, opened 2019-10-24
Fix warnings

emartynov:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring

## :scroll: Description

Fix warious compile warning

I like to have the compilation log without warnings.

Compile and run the application.

## :pencil: Checklist

## :crystal_ball: Next steps
Not all warnings are fixed but rest will require some rethinking.

Partial fix for #745


emartynov:
I experienced a couple of unsuccessful builds but message was not clear - gradle was not able to connect to child process:

```
Unable to connect to the child process 'Gradle Test Executor 8'.
  It is likely that the child process have crashed - please find the stack trace in the build log.
  This exception might occur when the build machine is extremely loaded.
  The connection attempt hit a timeout after 120.0 seconds (last known process state: STARTED, running: true).
```

I wonder if it is connected to my changes or just circle ci is busy.

florina-muntenescu:
could you fix the conflicts, please?

emartynov:
@florina-muntenescu done!

core/src/main/java/io/plaidapp/core/data/DataManager.kt
```diff
@@ -82,7 +83,9 @@ class DataManager @Inject constructor(
     init {
         sourcesRepository.registerFilterChangedCallback(filterListener)
         // build a map of source keys to pages initialized to 0
-        pageIndexes = sourcesRepository.getSourcesSync().map { it.key to 0 }.toMap().toMutableMap()
+        runBlocking {
```
florina-muntenescu: I'd rather not use `runBlocking` in non-tests code. Given that we want to re-write this at one point, let's leave this as it was, please

emartynov: Rebased my fork on latest changes and returned warning back as you asked

core/src/main/java/io/plaidapp/core/designernews/domain/LoadStoriesUseCase.kt
```diff
@@ -32,14 +31,14 @@ class LoadStoriesUseCase @Inject constructor(
 
     suspend operator fun invoke(page: Int): Result<List<Story>> {
```
ashdavies: I noticed this and was puzzled as I didn't think `Result` could be used as a return type?

florina-muntenescu: We are doing this in several places in Plaid.

PR #781, opened 2019-10-25
Fix some lint warning

emartynov:

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [x] Refactoring

## :scroll: Description

First banch of lint fixes

This is a showcase app. I think it should be an example of how to keep your project in good shape. 

Compile and run.

## :pencil: Checklist

## :crystal_ball: Next steps
Continue with lint report fixes.

Partial fix for #745

### DIFF
--- a/core/src/main/java/io/plaidapp/core/util/CollapsingTextHelper.java
+++ b/core/src/main/java/io/plaidapp/core/util/CollapsingTextHelper.java
@@ -74,6 +74,7 @@ public final class CollapsingTextHelper {
     private CharSequence mText;
     private CharSequence mTextToDraw;
     private boolean mIsRtl;
+    private boolean mUseTexture;
     private Bitmap mExpandedTitleTexture;
     private Paint mTexturePaint;
     private float mTextureAscent;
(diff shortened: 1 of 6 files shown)


core/src/main/java/io/plaidapp/core/util/CollapsingTextHelper.java
```diff
@@ -478,12 +459,6 @@ private void setInterpolatedTextSize(final float textSize) {
             }
             mIsRtl = calculateIsRtl(mTextToDraw);
         }
-        // Use our texture if the scale isn't 1.0
-        mUseTexture = USE_SCALING_TEXTURE && mScale != 1f;
```
florina-muntenescu: why was this removed?

emartynov: I thought that texture is only used on the older API and redundant since we target new API now. Let me know if this change was extra - I will restore the logic.

nickbutcher: This is taken from the old design lib, I'd recommend keeping it in sync with any changes MDC have made.

https://github.com/material-components/material-components-android/blob/master/lib/java/com/google/android/material/internal/CollapsingTextHelper.java

emartynov: Restored! Sorry!

emartynov: The keeping it in sync I would like to do in the different changes.

PR #792, opened 2019-11-26
Update Gradle wrapper to 5.6.4

friederbluemle:
This updates Gradle wrapper to `5.6.4`, including some changes to `gradle.properties` that bring the file closer to the default template. Note that the Gradle Daemon is enabled by default, so the following line is not needed:

```
org.gradle.daemon=true
```

### DIFF
--- a/gradle/wrapper/gradle-wrapper.properties
+++ b/gradle/wrapper/gradle-wrapper.properties
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-bin.zip
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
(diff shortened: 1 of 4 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,5 +1,5 @@
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.1-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.6.4-all.zip
```
keyboardsurfer: Please change back to `-bin`.

friederbluemle: @keyboardsurfer - Rebased and changed it back to `-bin`. Apologies for the slow reply!

Just curious: Is there any reason this project does not use the default `-all` distribution?
It's not only the default for new Android Studio projects, but also provides additional sources for debugging Gradle scripts. Also, since the Gradle wrapper distributions are **cached locally**, it's much more likely to save someone an additional download: Most Android projects use the default distribution, making them more likely to already exist on the machine.

friederbluemle:
Rebased onto latest default branch.

PR #806, opened 2019-12-09
Update ktlint and spotless versions

keyboardsurfer:
./gradlew spotlessApply

- [ ] Bugfix
- [ ] New feature
- [x] Enhancement
- [ ] Refactoring

## :pencil: Checklist

- [x] I added tests to verify changes
- [x] All tests passing


core/src/main/java/io/plaidapp/core/ui/widget/BadgedFourThreeImageView.kt
```diff
@@ -156,8 +156,8 @@ class BadgedFourThreeImageView(
 
         companion object {
             private const val GIF = "GIF"
-            private const val TEXT_SIZE = 12    // sp
-            private const val PADDING = 4       // dp
+            private const val TEXT_SIZE = 12 // sp
```
pfmaggi: I tend to prefer the other comment layout. Is this really what we want or is "just" the ktlint default?

keyboardsurfer: This is the ktlint default. For a comment suffix I'm okay with that default.

PR #823, opened 2020-03-02
Update Android Gradle plugin to 3.6.1

friederbluemle:
A small patch update to the Android Gradle plugin

https://androidstudio.googleblog.com/2020/02/android-studio-361-available.html


keyboardsurfer:
Thanks, every little helps. 😄

